PR #36, opened 2015-03-03
Convert the Notification API to use Bikeshed

beverloo:
This should be mostly similar to the Anolis version, minus the following differences:

- Typo fix: "create a notification" steps, step 5, +"option's".
- Typo fix: "create a notification" steps, step 8, s/notifications's/notification's/.
- Typo fix: ...when invoked, must return these steps  s/return/run/.

There are two problems I ran in to.

(1) IDL generation doesn't like the "any" data type. It doesn't insert a space between the property's type and name, e.g. "readonly attribute anydata;" and "anydata = null;".
(2) IDL generation fails on use of the "required" keyword in a dictionary. According to WebIRL, [this is allowed](http://heycam.github.io/webidl/#required-dictionary-member).

FATAL ERROR: IDL SYNTAX ERROR LINE: 20 - skipped: "required Notification notification"


beverloo:
+@TabAtkins FYI

beverloo:
Patched the widlparser for issue (2) here:
  https://github.com/plinss/widlparser/pull/5

beverloo:
Both issues in the widlparser have been fixed, and I merged in 11f25cc.

Please take a look :-).

beverloo:
dfn.js is being loaded as well now.

https://secure.peter.sh/notifications.html

beverloo:
Renamed "Version History" to commits, and moved it below participate.

While the generated file included in the PR is correct, it depends on the following Bikeshed PR:
  https://github.com/tabatkins/bikeshed/pull/377

PR #48, opened 2015-07-28
Add optional actions (buttons) to notifications

### LINKED ISSUE #44: Add optional action buttons to notifications
[Android 4.1+][1], [iOS 8+][2], [Chrome Apps][3] and [Mozilla XUL][4] all support showing action buttons at the bottom of a notification. This allows users to quickly handle the most common tasks for a particular notification, without having to open the originating application.

[1]: http://developer.android.com/design/patterns/notifications.html#actions
[2]: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIUserNotificationAction_class/
[3]: https://developer.chrome.com/apps/notifications#property-NotificationOptions-buttons
[4]: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Method/appendNotification

We'd like to add action buttons to the Notifications API, and implement them in Chrome.

Strawman API:
```webidl
dictionary NotificationAction {
    DOMString id;  // If clicked, this will become NotificationEvent.actionId
    DOMString title;  // Button text
}
partial dictionary NotificationOptions {
    sequence<NotificationAction> actions;
}
partial interface Notification {
    FrozenArray<NotificationAction> actions;
}
partial interface NotificationEvent {
    DOMString actionId;  // Id of clicked action, or null if user clicked notification body instead
}
```

Action support would be feature detected via `Notification.prototype.hasOwnProperty('actions')`. UAs which expose that must support up to 2 actions. If more than 2 actions are provided in `NotificationOptions.actions`, throw a TypeError. The number 2 was chosen because it's the maximum supported by iOS and Chrome OS, and although Android supports 3 buttons, Chrome for Android already uses one of the available slots to show a link to per-origin settings. (If later UAs want to support more actions, we could add a `maxSupportedActions` constant to Notification).

Presentation of actions would be implementation-defined, for example on Android action buttons are only shown when the notification is [expanded](http://developer.android.com/design/patterns/notifications.html#ExpandedLayouts). Hence websites should not rely on actions, and should provide the same functionality via alternate UI that can be reached by clicking on the notification body.

I explicitly left out action icons, since only Android supports them, and Android currently only supports icon resources shipped in a native app's apk (not icons downloaded at runtime), so they wouldn't be implementable yet there either.

@sicking @beverloo

johnmellor:
Resolves #44

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -100,6 +100,17 @@ associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
 <p class=note>Developers are encouraged to not convey information through an icon, sound,
 or vibration pattern that is not otherwise accessible to the end user.
 
+<p>A <a lt="concept notification">notification</a> has an associated list of
+zero or more <dfn>actions</dfn>. Each action has an associated
+<dfn lt="action title">title</dfn> and <dfn>action name</dfn>. Users may
+activate actions, as alternatives to activating the notification itself. The
+user agent must determine the <dfn>maximum number of actions</dfn> supported,
+within the constraints of the notification platform.
+
+<p class=note>Since display of actions is platform-dependent, developers are
+encouraged to make sure that any action a user can invoke from a notification is
+also available within the web application.
+
 <p>A <dfn>non-persistent notification</dfn> is a
 <a lt="concept notification">notification</a> without an associated
 <a>service worker registration</a>.
@@ -112,12 +123,24 @@ or vibration pattern that is not otherwise accessible to the end user.
 
 <hr>
 
-<p>To <dfn>create a notification</dfn>, given a <var>title</var> and
-<var>options</i>, run these steps:
+<p>To <dfn>create a notification</dfn>, given a <var>title</var>,
+<var>options</var>, and optionally a <var>serviceWorkerRegistration</var>, run
+these steps:
 
 <ol>
   <li><p>Let <var>notification</var> be a new
   <a lt="concept notification">notification</a>.
+
+  <li><p>If a <var>serviceWorkerRegistration</var> was provided, set
+  <var>notification</var>'s <a>service worker registration</a> to
+  <var>serviceWorkerRegistration</var>.
+
+  <li><p>If a <var>serviceWorkerRegistration</var> was not provided and
+  <var>options</var>'s <code>actions</code> is not empty, <a>throw</a> a
+  <code>TypeError</code> exception.
+
+  <p class=note><a>Actions</a> are only currently supported for
+  <a>persistent notifications</a>.
 
   <li><p>If <var>options</var>'s <code>silent</code> is true, and either
   <var>options</var>'s <code>sound</code> is present or <var>options</var>'s
@@ -189,6 +212,24 @@ or vibration pattern that is not otherwise accessible to the end user.
   <li><p>If <var>options</var>'s <code>sticky</code> is true, set
   <var>notification</var>'s <a>sticky preference flag</a>.
 
+  <li><p>Set <var>notification</var>'s list of <a>actions</a> to an empty list,
+  then for each <var>entry</var> in <var>options</var>'s <code>actions</code>,
+  up to the <a>maximum number of actions</a> supported (skip any excess
+  entries), perform the following steps:
+
+  <ol>
+    <li><p>Let <var>action</var> be a new <a lt="actions">action</a>.
+
+    <li><p>Set <var>action</var>'s <a>action name</a> to the <var>entry</var>'s
+    <code>action</code>.
+
+    <li><p>Set <var>action</var>'s <a lt="action title">title</a> to the
+    <var>entry</var>'s <code>title</code>.
+
+    <li><p>Append <var>action</var> to <var>notification</var>'s list of
+    <a>actions</a>.
+  </ol>
+
   <li><p>Return <var>notification</var>.
 </ol>
 
@@ -253,15 +294,23 @@ section of HTML. [[!HTML]]
 
 <p>User agents are expected to honor the Unicode semantics of the text of a
 <a lt="concept notification">notification</a>'s
-<a lt="concept title">title</a> and <a>body</a>. Each is expected to be
-treated as an independent set of one or more bidirectional algorithm paragraphs
-when displayed, as defined by the bidirectional algorithm's rules P1, P2, and
-P3, including, for instance, supporting the paragraph-breaking behavior of
-U+000A LINE FEED (LF) characters. For each paragraph of the
-<a lt="concept title">title</a> and <a>body</a>, the
+<a lt="concept title">title</a>, <a>body</a>, and the
+<a lt="action title">title</a> of each of its <a>actions</a>. Each is expected
+to be treated as an independent set of one or more bidirectional algorithm
+paragraphs when displayed, as defined by the bidirectional algorithm's rules P1,
+P2, and P3, including, for instance, supporting the paragraph-breaking behavior
+of U+000A LINE FEED (LF) characters. For each paragraph of the
+<a lt="concept title">title</a>, <a>body</a> and the
+<a lt="action title">title</a> of each of the <a>actions</a>, the
 <a lt="concept notification">notification</a>'s
-<a lt="concept direction">direction</a> provides the higher-level override
-of rules P2 and P3 if it has a value other than "<code>auto</code>". [[!BIDI]]
+<a lt="concept direction">direction</a> provides the higher-level override of
+rules P2 and P3 if it has a value other than "<code>auto</code>". [[!BIDI]]
+
+<p>The <a lt="concept notification">notification</a>'s
+<a lt="concept direction">direction</a> also determines the relative order in
+which the <a lt="concept notification">notification</a>'s <a>actions</a> should
+be displayed to the user, if the notification platform displays them side by
+side.
 
 
 <h3 id=language>Language</h3>
@@ -272,9 +321,11 @@ of rules P2 and P3 if it has a value other than "<code>auto</code>". [[!BIDI]]
 <p>The <a lt="concept notification">notification</a>'s
 <a lt="concept language">language</a> specifies the primary language for the
 <a lt="concept notification">notification</a>'s
-<a lt="concept title">title</a> and <a>body</a>. Its value is a string. The empty string
-indicates that the primary language is unknown. Any other string must be interpreted as a
-language tag. Validity or well-formedness are not enforced. [[!LANG]]
+<a lt="concept title">title</a>, <a>body</a> and the
+<a lt="action title">title</a> of each of its <a>actions</a>. Its value is a
+string. The empty string indicates that the primary language is unknown. Any
+other string must be interpreted as a language tag. Validity or well-formedness
+are not enforced. [[!LANG]]
 
 <p class=note>Developers are encouraged to only use valid language tags.
 
@@ -343,20 +394,26 @@ language tag. Validity or well-formedness are not enforced. [[!LANG]]
 
 <h3 id=activating-a-notification>Activating a notification</h3>
 
-<p>When a <a lt="concept notification">notification</a> <var>notification</var>
-is activated by the user, assuming the underlying notification platform supports
-activation, the user agent must (unless otherwise specified) run these steps:
+<p>When a <a lt="concept notification">notification</a> <var>notification</var>,
+or one of its <a>actions</a>, is activated by the user, assuming the underlying
+notification platform supports activation, the user agent must (unless otherwise
+specified) run these steps:
 
 <ol>
   <li><p>If <var>notification</var> is a <a>persistent notification</a>, run
   these substeps:
 
   <ol>
+    <li><p>Let <var>action</var> be the empty string.
+
+    <li><p>If one of <var>notification</var>'s <a>actions</a> was activated by
+    the user, set <var>action</var> to that action's <a>action name</a> string.
+
     <li><p>Let <var>callback</var> be an algorithm that when invoked with a
     <var>global</var>, <a lt="fire a service worker notification event named e">
     fires a service worker notification event</a> named
-    <code>notificationclick</code> given <var>notification</var> on
-    <var>global</var>.
+    <code>notificationclick</code> given <var>notification</var> and
+    <var>action</var> on <var>global</var>.
 
     <li><p>Then run <a>Handle Functional Event</a> with
     <var>notification</var>'s <a>service worker registration</a> and
@@ -453,6 +510,8 @@ interface Notification : EventTarget {
   static readonly attribute NotificationPermission permission;
   [Exposed=Window] static Promise&lt;NotificationPermission> requestPermission(optional NotificationPermissionCallback deprecatedCallback);
 
+  static readonly attribute unsigned long maxActions;
+
   attribute EventHandler onclick;
   attribute EventHandler onerror;
 
@@ -471,6 +530,7 @@ interface Notification : EventTarget {
   readonly attribute boolean requireInteraction;
   readonly attribute boolean sticky;
   [SameObject] readonly attribute any data;
+  [SameObject] readonly attribute FrozenArray&lt;NotificationAction> actions;
 
   void close();
 };
@@ -490,6 +550,7 @@ dictionary NotificationOptions {
   boolean requireInteraction = false;
   boolean sticky = false;
   any data = null;
+  sequence&lt;NotificationAction> actions = [];
 };
 
 enum NotificationPermission {
@@ -502,6 +563,11 @@ enum NotificationDirection {
   "auto",
   "ltr",
   "rtl"
+};
+
+dictionary NotificationAction {
+  required DOMString action;
+  required DOMString title;
 };
 
 callback NotificationPermissionCallback = void (NotificationPermission permission);
@@ -533,7 +599,7 @@ these steps:
   <li><p>If <var>options</var>'s <code>sticky</code> is present, <a>throw</a>
   a <code>TypeError</code> exception.
 
-  <li><p>If <a>entry settings object</a>'s <a>global object</a> is a
+  <li><p>If the <a>entry settings object</a>'s <a>global object</a> is a
   {{ServiceWorkerGlobalScope}} object, <a>throw</a> a
   <code>TypeError</code> exception.
 
@@ -564,7 +630,7 @@ these steps:
 <h3 id=static-members>Static members</h3>
 
 <p>The static <dfn attribute dfn-for=Notification><code>permission</code></dfn>
-attribute's getter must return <a>permission</a> for
+attribute's getter must return <a>permission</a> for the
 <a>entry settings object</a>'s <a>origin</a>.
 
 <p class=note>If you edit standards please refrain from copying the above. Synchronous
@@ -594,7 +660,7 @@ method, when invoked, must run these steps:
     <p><a>Queue a task</a> to run these subsubsteps:
 
     <ol>
-     <li><p>Set <a>permission</a> for <a>entry settings object</a>'s <a>origin</a> to
+     <li><p>Set <a>permission</a> for the <a>entry settings object</a>'s <a>origin</a> to
      <var>permission</var>.
 
      <li><p>If <var>deprecatedCallback</var> is given, invoke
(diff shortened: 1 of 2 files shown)


johnmellor:
[HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2F90e29647469fb26b1f707cf93d8c787b92d2bbdc%2Fnotifications.html&doc2=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2Fa74688eb593d6040173a77907879db47b96ce3c1%2Fnotifications.html) to aid reviewing.

notifications.bs
```diff
@@ -83,6 +83,14 @@ associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
 <p class=note>Developers are encouraged to not convey information through an icon, sound,
 or vibration pattern that is not otherwise accessible to the end user.
 
+<p>A <a lt="concept notification">notification</a> has an associated list of
+<dfn>actions</dfn> (which may be empty). Platforms may render these as buttons.
```
beverloo: nit: You could remove the parenthesis by rephrasing as "... zero or more actions."

notifications.bs
```diff
@@ -83,6 +83,14 @@ associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
...
+Users may activate these, as alternatives to activating the notification itself.
+
+<p class=note>Since action display is platform-dependent, developers are
```
annevk: "display of actions"

notifications.bs
beverloo: "Platforms may render these as buttons."

Let's stick to normative text in this paragraph, and move the display suggestions to the note below. This sentence is more like an example anyway.

annevk: I prefer "action list" as single `<dfn>`. And then "consisting of zero or more actions".

notifications.bs
```diff
@@ -227,16 +241,22 @@ section of HTML. [[!HTML]]
...
-U+000A LINE FEED (LF) characters. For each paragraph of the
-<a lt="concept title">title</a> and <a>body</a>, the
-<a lt="concept notification">notification</a>'s
+<a lt="concept title">title</a>, <a>body</a> and each of the <a>actions</a>.
+Each is expected to be treated as an independent set of one or more
+bidirectional algorithm paragraphs when displayed, as defined by the
+bidirectional algorithm's rules P1, P2, and P3, including, for instance,
+supporting the paragraph-breaking behavior of U+000A LINE FEED (LF) characters.
+For each paragraph of the <a lt="concept title">title</a>, <a>body</a> and each
+of the <a>actions</a>, the <a lt="concept notification">notification</a>'s
```
beverloo: Just the action's "title" attribute, BiDi doesn't have to care about the "action" value.

notifications.bs
```diff
@@ -426,6 +456,8 @@ interface Notification : EventTarget {
   static readonly attribute NotificationPermission permission;
   [Exposed=Window] static void requestPermission(optional NotificationPermissionCallback callback);
 
+  static readonly attribute unsigned long maxActions;
```
annevk: We generally don't expose limits. See e.g. vibration pattern.

notifications.bs
```diff
@@ -728,10 +800,12 @@ partial interface ServiceWorkerRegistration {
  Exposed=ServiceWorker]
 interface NotificationEvent : ExtendableEvent {
   readonly attribute Notification notification;
+  readonly attribute DOMString? action;
 };
 
 dictionary NotificationEventInit : ExtendableEventInit {
   required Notification notification;
+  DOMString? action = null;
```
annevk: I would prefer special casing the empty string. That would match `tag` better iirc.

annevk:
Thank you for this PR. I left some comments.

notifications.bs
```diff
@@ -246,9 +266,10 @@ of rules P2 and P3 if it has a value other than "<code>auto</code>". [[!BIDI]]
 <p>The <a lt="concept notification">notification</a>'s
 <a lt="concept language">language</a> specifies the primary language for the
 <a lt="concept notification">notification</a>'s
-<a lt="concept title">title</a> and <a>body</a>. Its value is a string. The empty string
-indicates that the primary language is unknown. Any other string must be interpreted as a
-language tag. Validity or well-formedness are not enforced. [[!LANG]]
+<a lt="concept title">title</a>, <a>body</a> and <a>actions</a>. Its value is a
```
beverloo: Again: Just the titles of the actions.

notifications.bs
beverloo: @sicking also suggested this in #44. Availability of actions might influence how a page emphasizes alternatives to actions on the page displayed in response to activating the notification.

That said, I'm having mixed feelings about only storing `maxActions` actions in the action list.

notifications.bs
```diff
@@ -647,20 +690,49 @@ getter must return the <a lt="concept notification">notification</a>'s
...
+objects dispatch events during their lifecycle, which developers can use to
+generate desired behaviors.
 
 <p>The <code>click</code> event dispatches when the user activates a
 notification.
 
 <pre class=example>
-var not = new Notification("Gebrünn Gebrünn by Paul Kalkbrenner", { icon: "newsong.svg", tag: "song" })
-notification.onclick = function() { displaySong(this) }</pre>
+var not = new Notification("Gebrünn Gebrünn by Paul Kalkbrenner", { icon: "newsong.svg", tag: "song" });
```
beverloo: s/.svg/.png/. At least in Chrome we don't support SVG images, and I imagine other implementations will also have difficulties with this.

annevk: You should though, so I don't support changing that (not as part of this PR anyway).

beverloo: Oh, I see that John's not actually changing the line. Let's leave that for a different issue :).

(The spec doesn't say anything about supported image formats. Supporting SVG will be rather complicated for us.)

annevk:
@johnmellor I'd also appreciate if you could file that IDL issue. The specification prose should operate on IDL primitives and ideally not call directly into JavaScript.

johnmellor:
Thanks! I've addressed all comments till now, except that I left in `maxActions` since there doesn't seem to be consensus there yet.

I filed the IDL issue as https://www.w3.org/Bugs/Public/show_bug.cgi?id=29004

notifications.bs
annevk: I mostly care about a consistent story. Why is this different from vibrate?

johnmellor: Historically, the DAP WG had fingerprinting concerns around exposing details of hardware vibrators.

By contrast, number of actions supported has much less fingerprinting risk, since it's only based on screen size and possibly OS, both of which are already exposed. Indeed Blink currently intends to always support 2 actions (both on Android and desktop).

Also, our thinking on fingerprinting has evolved, based on [data showing that it's infeasible to prevent fingerprinting](https://www.chromium.org/Home/chromium-security/security-faq#TOC-Why-isn-t-passive-browser-fingerprinting-including-passive-cookies-in-Chrome-s-threat-model-).

annevk: But we are exposing the vibrate limit, just in a different way.

johnmellor:
I've updated this PR, in particular:

- I realised that I can't share NotificationEvent with non-persistent
notifications, since it inherits from ExtendableEvent. Additionally, it's
unclear whether actions make sense on non-persistent notifications, which are
more toast-like. Hence actions can now only be provided for persistent
notifications. Passing actions for non-persistent notifications throws a
TypeError (though perhaps it would be more forward-compatible to do nothing?).

- I tried to explain the freezing of the actions attribute more precisely (based on
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29004, it seems that using a
FrozenArray is still a reasonable choice, and freezing the values in the
FrozenArray doesn't seem totally crazy, though it remains unclear whether
there will be a WebIDL primitive for this).

- I rebased and ran Bikeshed.

- (I haven't changed maxActions, since I haven't heard any recent opinions for or
against).

[Updated HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2F90e29647469fb26b1f707cf93d8c787b92d2bbdc%2Fnotifications.html&doc2=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2F316a115f59a15e68150c624648d0453dfcd6deab%2Fnotifications.html)

Please take another look - thanks!

notifications.bs
```diff
@@ -189,6 +209,10 @@ or vibration pattern that is not otherwise accessible to the end user.
   <li><p>If <var>options</var>'s <code>sticky</code> is true, set
   <var>notification</var>'s <a>sticky preference flag</a>.
 
+  <li><p>Set <var>notification</var>'s list of <a>actions</a> to the first
+  {{Notification/maxActions}} entries from <var>options</var>'s
```
annevk: maxActions will have to be some concept if we want to use it like this. Referencing code directly doesn't work well since it can be overridden by script.

notifications.bs
```diff
@@ -253,15 +277,23 @@ section of HTML. [[!HTML]]
...
 <p>User agents are expected to honor the Unicode semantics of the text of a
 <a lt="concept notification">notification</a>'s
-<a lt="concept title">title</a> and <a>body</a>. Each is expected to be
-treated as an independent set of one or more bidirectional algorithm paragraphs
-when displayed, as defined by the bidirectional algorithm's rules P1, P2, and
-P3, including, for instance, supporting the paragraph-breaking behavior of
-U+000A LINE FEED (LF) characters. For each paragraph of the
-<a lt="concept title">title</a> and <a>body</a>, the
+<a lt="concept title">title</a>, <a>body</a>, and the
+{{NotificationAction/title}} of each of its <a>actions</a>. Each is expected to
```
annevk: Same problem here.

notifications.bs
```diff
@@ -997,6 +1091,8 @@ urlPrefix: http://www.w3.org/TR/vibration/
     text: validate and normalize
   urlPrefix: #idl-def-; type: interface
     text: VibratePattern; url: VibratePattern
+urlPrefix: https://people.mozilla.org/~jorendorff/es6-draft.html#sec-object.; type: dfn
```
annevk: We should probably use https://tc39.github.io/ecma262/ now. @domenic?

notifications.bs
johnmellor: Done.

notifications.bs
johnmellor: Done, though note that it made things much more verbose, for example rather than:
```html
  <li><p>Set <var>notification</var>'s list of <a>actions</a> to the first
  <var>n</var> entries from <var>options</var>'s <code>actions</code>, where
  <var>n</var> is the <a>maximum number of actions</a> supported.
```
I ended up with:
```html
  <li><p>Set <var>notification</var>'s list of <a>actions</a> to an empty list,
  then for each <var>entry</var> in <var>options</var>'s <code>actions</code>,
  up to the <a>maximum number of actions</a> supported (skip any excess
  entries), perform the following steps:

  <ol>
    <li><p>Let <var>action</var> be a new <a lt="actions">action</a>.

    <li><p>Set <var>action</var>'s <a>action name</a> to the <var>entry</var>'s
    <code>action</code>.

    <li><p>Set <var>action</var>'s <a lt="action title">title</a> to the
    <var>entry</var>'s <code>title</code>.

    <li><p>Append <var>action</var> to <var>notification</var>'s list of
    <a>actions</a>.
  </ol>
```
And similarly for the conversion back from concept to `NotificationAction`.

notifications.bs
johnmellor: Done.

johnmellor:
[Updated HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2F90e29647469fb26b1f707cf93d8c787b92d2bbdc%2Fnotifications.html&doc2=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2Fd3aa9b127e8a47ba9fb136649b981874ed3cabf5%2Fnotifications.html)

johnmellor:
Re-added explicit mention of creating a frozen array, to avoid confusion. [Updated HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2Ff72a0417b9cacc3e29194bc7e0e602d97224e5a5%2Fnotifications.html&doc2=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2Ffe9c3cce079fbdcf7d764c09ff829ee3f1d29f25%2Fnotifications.html).

notifications.bs
```diff
@@ -100,6 +100,17 @@ associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
 <p class=note>Developers are encouraged to not convey information through an icon, sound,
 or vibration pattern that is not otherwise accessible to the end user.
 
+<p>A <a lt="concept notification">notification</a> has an associated list of
+zero or more <dfn>actions</dfn>. Each action has an associated
+<dfn lt="action title">title</dfn> and <dfn>action name</dfn>. Users may
+activate actions, as alternatives to activating the notification itself. The
+user agent should determine the <dfn>maximum number of actions</dfn> supported,
```
annevk: Must, no? Because if it doesn't, what would `maxActions` return?

notifications.bs
```diff
@@ -343,20 +394,26 @@ language tag. Validity or well-formedness are not enforced. [[!LANG]]
...
 
 <ol>
   <li><p>If <var>notification</var> is a <a>persistent notification</a>, run
   these substeps:
 
   <ol>
+    <li><p>If one of <var>notification</var>'s <a>actions</a> was activated by
+    the user, let <var>action</var> be that action's <a>action name</a> string.
+
+    <li><p>Otherwise, let <var>action</var> be the empty string.
```
annevk: I think I would prefer reversing these two steps and explicitly setting setting `<var>action</var>` if an action happened to be activated. Using "let" twice feels somewhat unclean (although I realize it's currently common in the HTML standard).

notifications.bs
```diff
@@ -471,6 +530,7 @@ interface Notification : EventTarget {
   readonly attribute boolean requireInteraction;
   readonly attribute boolean sticky;
   [SameObject] readonly attribute any data;
+  readonly attribute FrozenArray&lt;NotificationAction> actions;
```
annevk: This should be `[SameObject]`, no?

notifications.bs
annevk: Could you perhaps not modify this example as you are not changing it? Same for the next line. Omitting `;` is perfectly fine.

notifications.bs
```diff
@@ -704,20 +773,75 @@ getter must return the <a lt="concept notification">notification</a>'s
...
+  <ol>
+    <li><p>Let <var>action</var> be a new {{NotificationAction}}.
+
+    <li><p>Set <var>action</var>'s {{NotificationAction/action}} to
+    <var>entry</var>'s <a>action name</a>.
+
+    <li><p>Set <var>action</var>'s {{NotificationAction/title}} to
+    <var>entry</var>'s <a lt="action title">title</a>.
+
+    <li><p>Call <a lt="freeze">Object.freeze</a> on <var>action</var>.
```
annevk: I think you need to say here that you are using the initial value of `Object.freeze` so it cannot possibly be overridden from script. There should probably also be some XXX type of comment in the source here and a bug on IDL since it's somewhat strange to invoke JavaScript on what are effectively IDL constructs.

notifications.bs
johnmellor: Done.

notifications.bs
johnmellor: Done.

notifications.bs
johnmellor: Done (indeed, there's currently no reason the UA would replace actions with a different FrozenArray).

notifications.bs
johnmellor: Done (I've added an XXX comment, which links to https://www.w3.org/Bugs/Public/show_bug.cgi?id=29004, and explained in the prose that this is "to prevent accidental mutation by scripts").

johnmellor:
[Updated HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2Ff72a0417b9cacc3e29194bc7e0e602d97224e5a5%2Fnotifications.html&doc2=https%3A%2F%2Frawgit.com%2Fjohnmellor%2Fnotifications%2F5022b8399443601162f77557d023e80411a15154%2Fnotifications.html)

notifications.bs
johnmellor: The example used inconsistent variable names, `var not` then `notification.onclick`. That seems worth fixing, though I'm happy to do that as a separate PR if you'd prefer? Left it in for now.

notifications.bs
```diff
@@ -189,6 +212,24 @@ or vibration pattern that is not otherwise accessible to the end user.
   <li><p>If <var>options</var>'s <code>sticky</code> is true, set
   <var>notification</var>'s <a>sticky preference flag</a>.
 
+  <li><p>Set <var>notification</var>'s list of <a>actions</a> to an empty list,
```
annevk: Sorry for not spotting this earlier. Since the `<p>` is not the only child of the `<li>` it needs to appear on its own line and indented. Same for the `<ol>` (and its children).

notifications.bs
```diff
@@ -704,20 +773,81 @@ getter must return the <a lt="concept notification">notification</a>'s
...
 <a lt="concept notification">notification</a>'s <a>data</a>.
 
+<p>The <dfn attribute dfn-for=Notification><code>actions</code></dfn>
+attribute's getter must return the result of the following steps:
+
+<ol>
+  <li><p>Let <var>frozenActions</var> be an empty list of type
+  {{NotificationAction}}.
+
+  <li><p>For each <var>entry</var> in the
```
annevk: Same issue with multiple children here.

notifications.bs
annevk: I see. That seems worth fixing. I guess I can live with the added semicolons.

annevk:
@sicking, does this look good to you too?

notifications.bs
johnmellor: Hmm, the prevailing style seems to be not to indent these - see [line 289][1],  [line 306][2],  [line 351][3],  [line 369][4], and  [line 413][5].

[1]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L289
[2]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L306
[3]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L351
[4]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L369
[5]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L413

The only examples I can find of indenting the children - [line 434][6],  [line 581][7], and  [line 594][8] - all use single-space indent for the entire enclosing `<ol>`. But I'd rather not convert other `<ol>`s like the giant [create-a-notification](https://notifications.spec.whatwg.org/#create-a-notification) `<ol>` to single-space indent.

[6]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L434
[7]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L581
[8]: https://github.com/whatwg/notifications/blob/f72a0417b9cacc3e29194bc7e0e602d97224e5a5/notifications.bs#L594

Do still you want me to indent these? And if so, shall I use 2-space indent even though there's no precedent for that?

annevk: Hmm, I guess it's fine to leave this for now then. Thank you for the research. Should probably be cleaned up at some point, but given the precedent it's not really important to do it as part of this.

johnmellor:
(I reordered the `<li>`s I added to the `create a notification` algorithm, since it's [easier to implement](https://codereview.chromium.org/1409143004/diff/1/third_party/WebKit/Source/modules/notifications/Notification.cpp#newcode82) when the steps that only apply to non-persistent notifications aren't interleaved with the parts of the algorithm that apply to both).

sicking:
I looked through the HTML-diff, mainly looking at the WebIDL changes, and this looks great to me!

PR #54, opened 2015-10-02
Add Notification.timestamp and NotificationOptions.timestamp

### LINKED ISSUE #20: Add date/timestamp field
See http://lists.w3.org/Archives/Public/public-whatwg-archive/2014Aug/0017.html

@sicking, I guess this should just mimic the timestamp used for Date objects? Do we expect applications to worry about timezones?

beverloo:
Fixes #20.

The `timestamp` of a notification is the time, in milliseconds since
the epoch, of the event for which the notification was created. Web
developers can use this if the time of this event does not match the
time at which the notification is being shown, which can be the case
when notifying the user of upcoming calendar events.

Note that the `Boilerplate` change in the meta-data is Tab's
suggestion in https://github.com/tabatkins/bikeshed/issues/470.

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -7,6 +7,7 @@ Status: LS
 No Editor: true
 Abstract: This standard defines an API to display notifications to the end user, typically outside the top-level browsing context's viewport. It is designed to be compatible with existing notification systems, while remaining platform-independent.
 Logo: https://resources.whatwg.org/logo-notifications.svg
+Boilerplate: omit feedback-header
 !Participate: <a href=https://github.com/whatwg/notifications>GitHub whatwg/notifications</a> (<a href=https://github.com/whatwg/notifications/issues/new>new issue</a>, <a href=https://github.com/whatwg/notifications/issues>open issues</a>)
 !Participate: <a href=https://wiki.whatwg.org/wiki/IRC>IRC: #whatwg on Freenode</a>
 !Commits: <a href="https://github.com/whatwg/notifications/commits">GitHub whatwg/notifications/commits</a>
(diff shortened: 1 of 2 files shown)


annevk:
@sicking, could you review?

sicking:
Just two things:

I think we should have an informative not that says that the timestamp can both be in the future, for example for notifications about a meeting starting in 10 minutes, or in the past, when the phone regains internet connectivity and a messaging app detects that a message was sent to the user an hour ago.

I think we should also default the timestamp to the current time if no timestamp is provided?

beverloo:
I have added a \<p class=note> in the list of definitions. Does this describe the intention well enough?

The `timestamp` member in the `NotificationOptions` dictionary has no default value. The behavior is explained in the  "create a notification" steps:

> If options’s timestamp is present, set notification’s timestamp to the value. Otherwise,
> set notification’s timestamp to the number of milliseconds that passed since 00:00:00
> UTC on 1 January 1970.

beverloo:
@sicking Friendly ping :) WDYT?

sicking:
Sorry, didn't realize this was waiting on me. Looks good to me. Only minor thing might be to clarify the point in time at which the number of milliseconds is measured. I.e. something like:

Otherwise, set notification’s timestamp to the number of milliseconds that passed between 00:00:00
UTC on 1 January 1970 and `Notification` constructor was called.

But I'm really fine either way.

annevk:
One nit I have is to remove the trailing dot from the first line of the commit message. We're treating that as a title of sorts these days and those don't have dots.

beverloo:
Both done. Thank you!

Issue #31, opened 2015-02-25
Notification close event for a notification


gauntface:
At some point there was a close event proposed for Notifications (https://developer.mozilla.org/en-US/docs/Web/API/Notification/onclose) which no longer seems to be in the spec.

It would be useful to have this event for stats and ultimately shaping when developers send notifications (you can imagine a scenario of realizing that send push messages on specific categories are always closed by the user so the site reduces the number of pushes for that category).

Is there a reason this isn't supported by the spec?

annevk:
Judging from https://github.com/whatwg/notifications/commit/53ebcf25458f7581570a228cfa11ad7cb27ff4ea I suspect the October 2014 archives and perhaps the month prior of the WHATWG mailing list have definitive answers.

The main reason it's no longer there is that non-service worker notifications are no longer persistent and disappear after a few seconds. And for service worker notifications the use cases were not compelling enough.

marco-c:
It could be useful to gather statistics about the notifications. For example, I might want to know if the push notifications I'm sending are effective or not (I can count the number of notifications clicked and the number of notifications ignored, but I can't count the number of notifications closed).

annevk:
@marco-c see #61, we're considering adding this back for persistent notifications.

PR #61, opened 2016-01-25
Add notificationclose event

nicolassatragno:
This PR adds a `notificationclose` event to the service worker API, fired for persistent notifications closed by the user. This should satisfy #31 and marco-c/wp-web-push#63.

Right now, the *action* attribute would be set to null. As an alternative, we might want to split the `Notification` object into two different objects: `NotificationClick` and `NotificationClose`.

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -457,6 +457,21 @@ must be run.
   <li><p>If <var>notification</var> is not in the <a>list of notifications</a>,
   terminate these steps.
 
+  <li><p>If <var>notification</var> is a <a>persistent notification</a> and
+  <var>notification</var> was closed by the user, run these substeps:
+
+  <ol>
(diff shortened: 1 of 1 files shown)


annevk:
@beverloo care to review? Looks good to me.

@nicolassatragno could you maybe merge your commits and force push?

annevk:
@nicolassatragno also, thank you for doing this! Much appreciated.

nicolassatragno:
@annevk squashed both commits and pushed. Thanks for the quick review :smile:

beverloo:
This looks good to me as well. Thank you! :)

marco-c:
Unfortunately for https://github.com/marco-c/wp-web-push/issues/63 I'd need an event when a non-persistent notification gets closed.

Issue #59, opened 2016-01-20
Add optional icon for NotificationAction


mvano:
On some platforms the button rendered for a NotificationAction can display an icon. [This example](http://arstechnica.com/gadgets/2014/06/android-wear-review/3/) shows it on an Android phone and watch.

I propose adding a simple icon attribute (like the one on [NotificationOptions](https://notifications.spec.whatwg.org/#dom-notificationoptions-icon)):
```
USVString icon;
```

annevk:
This seems reasonable. I wonder if any other vendors have implemented this feature.

johnmellor:
There are two subtleties here:

1. Authors may want to provide multiple resolutions/densities. But I think it's good to go ahead with a single URL string for now, then later on we can match whatever #5 and #28 decide to do for `NotificationOptions.icon` (presumably we'll either change both to be a union type, or extend the string parsing like srcset does).

2. The required foreground and background color of the icon are platform-specific (just on Android, KitKat and below show white on black icons, Lollipop and above show dark gray on light gray icons, and Android Wear shows white on blue icons). It's unreasonable to expect authors to provide different icons for every platform (or two icons at once for Lollipop+ phones connected to Android Wear), so instead I'd suggest specifying that the author SHOULD provide an image whose alpha channel can be used on its own as an alpha mask, such that UAs MAY ignore the color channels and generate the appropriate flat color icon(s) from the alpha mask.
For example some UAs would render the following icon:
[![colorful_settings_cog_checkerboard](https://cloud.githubusercontent.com/assets/563043/12455593/310c4a06-bf94-11e5-9ce4-599223c6c207.png)](https://cloud.githubusercontent.com/assets/563043/12455600/34771f7c-bf94-11e5-9c22-1ffe83987e50.png)
as:
![phone_notification](https://cloud.githubusercontent.com/assets/563043/12455585/2c8fc20a-bf94-11e5-849d-e3a504170481.png)

mvano:
On 1 I agree that multiple image definitions can be done as a next step.

On 2 I am less keen, though it is a good point. While providing an easily masked icon is good advice for the current crop of popular platforms, I'd rather not include such a recommendation in the normative text. Perhaps it's more appropriate in a non-normative note.

mvano:
>I wonder if any other vendors have implemented this feature.

[...]

Issue #28, opened 2014-12-14
Supporting multiple image definitions for a Notification


beverloo:
Currently Notifications support a single image to be defined, but on today's world of many different screen densities we may want to consider a syntax for multiple syntax.

My proposal is to adopt the Web Manifest's syntax for defining an icon size:
http://w3c.github.io/manifest/#icons-member

This would look as follows:

```
registration.showNotification('My Notification', {
  body: 'Hello, world!',
  icon: '/legacy-icon.png',  // optional, only used as fall-back
  icons: [
    { src: '/icon-1x.png',
      type: 'image/png',
      density: 1 },
    { src: '/icon-2x.png',
      density: 2 },
    { src: 'icon.ico',
      sizes: '128x128 256x256' }
  ]
});
```

On a tangent, we may also want to consider some sort of "small icon", to be used when the platform displays such icons in a status bar. (E.g. my [little flag man](https://secure.peter.sh/tests/screenshots/notifications-ux-icon-nov24.png) on Android -- we're going to show a generic Notification icon when shipping them.)

+@mounirlamouri

annevk:
@foolip @richtr Media Session has this problem as well. We should come up with a solution that works for both and elsewhere. Since manifests are still somewhat at risk I'd prefer us reusing a pattern of something that's implemented, like `<picture>`.

foolip:
Very interesting. Media Session will indeed need this, because you'll want to change the icon of the media playback notification. However, we also want to change the lockscreen background. Both of these are currently in `MediaMetadata.artwork`, but it would probably make sense to split them into `icon` and `artwork` or perhaps `background`.

It seems like https://github.com/whatwg/mediasession/issues/58 applies very much here as well. How do we expose the load of this kind of multi-image icon, and how do we make it possible to preload the icon using Fetch?

richtr:
@annevk wrote:
> I'd prefer us reusing a pattern of something that's implemented, like `<picture>`.

Do you mean letting [`icon`](https://notifications.spec.whatwg.org/#dom-notification-icon) also take [image candidate strings](https://html.spec.whatwg.org/multipage/embedded-content.html#parse-a-srcset-attribute) in the same way `srcset` works? I honestly prefer that idea over using the manifest syntax.

mounirlamouri:
Why is Manifest "at risk"? It is implemented by Edge, Firefox, Google and Opera and used in the wild.

mvano:
[...]

PR #62, opened 2016-01-28
Define icons for actions

mvano:
This PR adds an <code>icon</code> attribute to <code>NotificationAction</code>, as proposed in #59

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -113,12 +113,12 @@ notification platform.
 encouraged to make sure that any action a user can invoke from a notification is
 also available within the web application.
 
-<p class=note>On some platforms an <a>action icon resource</a> may be modified
-before being displayed to fit better with the platform's visual style. For
-example, it may be painted a specific color, corners may be rounded, a drop
-shadow may be applied. It is recommended to use an icon design that handles such
-modifications gracefully and does not lose important information through e.g.
-loss of color or clipped corners.
+<p class="note no-backref">Some platforms may modify an
+<a>action icon resource</a> to better match the platform's visual style before
+displaying it to the user, for example by rounding the corners or painting it in
+a specific color. It is recommended to use an icon that handles such cases
+gracefully and does not lose important information through e.g. loss of color or
+clipped corners.
 
 <p>A <dfn>non-persistent notification</dfn> is a
 <a lt="concept notification">notification</a> without an associated
@@ -516,7 +516,7 @@ must be run.
 <ol>
   <li><p>Wait for any <a lt="fetch">fetches</a> to complete and
   <var>notification</var>'s <a>icon resource</a> and <a>sound resource</a>
-  to be set (if any), as well as the <a>action icon resource</a>s for the
+  to be set (if any), as well as the <a>action icon resources</a> for the
   <var>notification</var>'s <a>actions</a> (if any).
 
   <li><p>Display <var>notification</var> on the device (e.g. by calling the
@@ -536,7 +536,7 @@ must be run.
 <ol>
   <li><p>Wait for any <a lt="fetch">fetches</a> to complete and
   <var>notification</var>'s <a>icon resource</a> and <a>sound resource</a>
-  to be set (if any), as well as the <a>action icon resource</a>s for the
+  to be set (if any), as well as the <a>action icon resources</a> for the
   <var>notification</var>'s <a>actions</a> (if any).
 
  <li><p>Replace <var>old</var> with <var>new</var>, in the same position, in the
@@ -1133,6 +1133,7 @@ Jonas Sicking,
 Michael Cooper,
 Michael Henretty,
 Michael™ Smith,
+Michael van Ouwerkerk,
 Nicolás Satragno,
 Olli Pettay,
 Peter Beverloo,
--- a/notifications.bs
+++ b/notifications.bs
@@ -102,14 +102,23 @@ or vibration pattern that is not otherwise accessible to the end user.
 
 <p>A <a lt="concept notification">notification</a> has an associated list of
 zero or more <dfn>actions</dfn>. Each action has an associated
-<dfn lt="action title">title</dfn> and <dfn>action name</dfn>. Users may
-activate actions, as alternatives to activating the notification itself. The
-user agent must determine the <dfn>maximum number of actions</dfn> supported,
-within the constraints of the notification platform.
+<dfn lt="action title">title</dfn> and <dfn>action name</dfn> and <em>can</em>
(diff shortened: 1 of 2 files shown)


notifications.bs
```diff
@@ -102,15 +102,24 @@ or vibration pattern that is not otherwise accessible to the end user.
...
 <p class=note>Since display of actions is platform-dependent, developers are
 encouraged to make sure that any action a user can invoke from a notification is
 also available within the web application.
 
+<p class=note>On some platforms an <a>action icon resource</a> may be modified
+before being displayed to fit better with the platform's visual style. For
+example, it may be painted a specific color, corners may be rounded, a drop
+shadow may be applied. It is recommended to use an icon design that handles such
+modifications gracefully and does not lose important information through e.g.
+loss of color or clipped corners.
```
beverloo: "may be modified" is a bit ambiguous. What about the following phrasing?

> Some platforms may modify an [action item resource] to better match the platform's visual style
before displaying it to the user, for example by adding rounded corners or painting it in a specific
color. It is recommended to use an icon that handles such cases gracefully and does not lose
important information through e.g. loss of color or clipped corners.

It would be great if we could add something among the lines of "TEST!", but it's already quite verbose as-is.

notifications.bs
```diff
@@ -102,15 +102,24 @@ or vibration pattern that is not otherwise accessible to the end user.
...
+<dfn>action icon resource</dfn>. Users may activate actions, as alternatives to
+activating the notification itself. The user agent must determine the
+<dfn>maximum number of actions</dfn> supported, within the constraints of the
+notification platform.
 
 <p class=note>Since display of actions is platform-dependent, developers are
 encouraged to make sure that any action a user can invoke from a notification is
 also available within the web application.
 
+<p class=note>On some platforms an <a>action icon resource</a> may be modified
```
beverloo: Since there's a reference in here, I think you should use `class="note no-backref"`.

notifications.bs
```diff
@@ -483,7 +516,8 @@ must be run.
 <ol>
   <li><p>Wait for any <a lt="fetch">fetches</a> to complete and
   <var>notification</var>'s <a>icon resource</a> and <a>sound resource</a>
-  to be set (if any).
+  to be set (if any), as well as the <a>action icon resource</a>s for the
```
beverloo: It's customary (at least in this spec) for plurals to be linked completely, using the `lt` attribute for the singular form.

> \<a lt="action icon resource">action icon resources\</a>

(dito on line 539)

beverloo:
Overall I think this change is good, a few comments. Two more points:

- How does this work with #28? I'd be OK with updating `Notification.icon` and `NotificationAction.icon` together (since we'll need to accept both USVString and an array anyway), but I do think this raises the priority of #28, especially since you add a note further about platform-specific behaviour.

- Please add yourself to the acknowledgements.

notifications.bs
mvano: Done.

notifications.bs
mvano: Done on rephrasing it a bit.

notifications.bs
mvano: As it turns out, Bikeshed seems to be clever about plurals, it just links them correctly.

mvano:
I think #28 can be addressed after getting action icons in. I agree that this PR helps motivate addressing it.

mvano:
All done.

beverloo:
This looks good to me, also based on the additional data Michael provided in #59. Let's see if we can get some functional feedback from Mozilla.

WDYT @annevk?

notifications.bs
```diff
@@ -102,15 +102,24 @@ or vibration pattern that is not otherwise accessible to the end user.
...
 
 <p class=note>Since display of actions is platform-dependent, developers are
 encouraged to make sure that any action a user can invoke from a notification is
 also available within the web application.
 
+<p class="note no-backref">Some platforms may modify an
+<a>action icon resource</a> to better match the platform's visual style before
+displaying it to the user, for example by rounding the corners or painting it in
+a specific color. It is recommended to use an icon that handles such cases
+gracefully and does not lose important information through e.g. loss of color or
```
annevk: There should be a comma before "e.g.".

annevk:
This seems reasonable to me. Still trying to find some additional folks to review new features, but this is not that big.

notifications.bs
mvano: While [this](http://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e) and [this](http://english.stackexchange.com/questions/7946/use-of-e-g-are-parentheses-necessary) page explain how this is recommended by some style guides, primarily American ones, it is not usually required. If anything, a comma after is more commonly recommended, and one could even use both one before and after.

 Personally, I find commas look unappealing in combination with the periods, and would prefer not to add any.

annevk: Yeah, maybe that's okay. @sideshowbarker has been asking me to do it, but thus far whenever I ask it of others they don't like it. Maybe we need a WHATWG style guide.

annevk:
I'll let @sideshowbarker have a look and then probably merge this tomorrow if I don't forget. Feel free to ping me. If you could rebase/squash that'd be appreciated. I'll try to add some Merge advice around the same time.

sideshowbarker:
I don’t feel strongly but among that arguments in favor or using comma after _e.g._ are:

* it’s never _incorrect_ no matter what style you follow (as opposed to just being seen as _unnecessary_)
* it’s the style used in other Web standards (which for consistency with one another mostly follow a convention of using American English style)

So I would prefer that either:

* if you want to use _e.g._, it should be followed with a comma, and we also make a practice of consistently using a comma after _e.g._ in all our specs
* you avoid the question by instead just writing it out as _for example_—with a comma after it and before it as needed; for example:

  > and does not lose important information through, for example, loss of color

Incidentally when it’s written out as _for example_ I think the purpose of the comma(s) is clear—in that, you wouldn’t write the above as:

> and does not lose important information through for example loss of color

mvano:
>Incidentally when it’s written out as for example I think the purpose of the comma(s) is clear
Oh the purpose is clear. But to me, "e.g." just expands to ", for example," with the commas included, as visually it looks similar enough with the periods.

Anyway, I'm not the editor of this spec, so if consensus is to insist on the commas just let me know whether to add one before, after, or both.

annevk:
Committed as https://github.com/whatwg/notifications/commit/362780b1e173d547955ac0b90c427a13bf16d530 with gratuitous use of commas.

Thank you!

PR #67, opened 2016-03-16
Fix #65: define badge

### LINKED ISSUE #65: Add small icon to Notification and NotificationOptions
I'd like to split off the small icon discussion from issue #28 where it was first suggested for this specification, but it was not the main topic of that issue.

If it is provided a small icon may be used, if supported by the platform, to represent the notification when there is not enough space to display the notification itself. It may also be displayed inside the notification.

A site might use the same small icon for each notification, in that case it is simply a site identifier. Alternatively, different icons may be used by a site to distinguish types of notifications, like text messages and picture messages. Or, it could be used to express that there is only one piece of information vs multiple.

This concept exists on multiple platforms:

* For [Android notifications](http://developer.android.com/guide/topics/ui/notifiers/notifications.html) it is actually a required part of a notification, so that the notification can be represented in the status bar. There, user agents must fall back to a default icon when no small icon is provided by the site.
* It is used in the Windows lock screen on both [desktop](https://msdn.microsoft.com/en-us/library/windows/apps/hh779720.aspx) and [mobile](https://msdn.microsoft.com/en-us/library/windows/apps/jj207048(v=vs.105).aspx).
* There may be more? Please let me know!

In various contexts this icon may be referred to as "small icon", "app icon", or "notification badge". I suggest calling it "small icon" in general, and "smallIcon" in IDL:

``` webidl
dictionary NotificationOptions {
  USVString smallIcon;
};
interface Notification {
  readonly attribute USVString smallIcon;
};
```

I think this also raises the priority of committing to a solution for multiple image definitions as discussed in issue #28.

mvano:


notifications.bs
```diff
@@ -98,8 +98,16 @@ associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
...
 <dfn>service worker registration</dfn>.
 
-<p class=note>Developers are encouraged to not convey information through an icon, sound,
-or vibration pattern that is not otherwise accessible to the end user.
+<p>A <a lt="concept notification">notification</a> <em>can</em> have an
+associated <dfn>badge URL</dfn> and <dfn>badge resource</dfn>. A <a>badge
+resource</a> <em>may</em> may be used to represent the
+<a lt="concept notification">notification</a> when there is not enough space to
+display the <a lt="concept notification">notification</a> itself. It may also be
+displayed inside the <a lt="concept notification">notification</a>.
```
annevk: I think we need a little bit more context here with respect to how this feature relates to icon.

mvano: I've added a bit more detail, let me know if that's the kind of thing you were looking for.

notifications.bs
```diff
@@ -103,7 +103,8 @@ associated <dfn>badge URL</dfn> and <dfn>badge resource</dfn>. A <a>badge
 resource</a> <em>may</em> may be used to represent the
 <a lt="concept notification">notification</a> when there is not enough space to
 display the <a lt="concept notification">notification</a> itself. It may also be
-displayed inside the <a lt="concept notification">notification</a>.
+displayed inside the <a lt="concept notification">notification</a>, but then it
+should have a lower visual priority than the <a>icon resource</a>.
```
annevk: Not sure, but s/lower/less/? Seems reasonable otherwise.

mvano: Sure, done.

annevk:
Okay, once squashed this seems okay to land, after @beverloo's question in #65 is answered.

mvano:
Thanks Anne. I squashed them.

annevk:
Thank you, landed as https://github.com/whatwg/notifications/commit/6ff593f7bc918f3ebbd87ffd2f5abf36a1346182.

PR #75, opened 2016-05-26
Fix some typos.

mvano:


annevk:
Thanks, landed as https://github.com/whatwg/notifications/commit/41843b5f65a9fb5b3d220a6deb4c1fb6d2f5d4f9.

annevk:
(I'll give you write access so you can create branches directly against the repository which should give cleaner merges when I need to generate the output. Master branch will still be protected though, FWIW.)

PR #76, opened 2016-06-01
Multiple icon definitions

mvano:
I think the initial patch has the general shape of what I want to do for #28.

I did find there's two ways we could go with the attribute getters on a notification, for example for the icon. To me it's not clear whether we need a single `icon` getter that reflects the selected url, or both `icon` and `icons` which reflect the exact input from options.

Previously `icon` reflected a serialization of the options `icon` input that was parsed with baseURL, not the exact original input. That means, for example, that if parsing fails, the getter returns the empty string. The implications for `NotificationIcon` are not clear to me: would invalid input such as for `type` or `sizes` need to be dropped in the same way? That might surprise developers.

***
<a href="https://whatpr.org/notifications/76.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (54dd556)">Preview</a> | <a href="https://whatpr.org/notifications/76/41843b5...54dd556.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (54dd556)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -617,7 +617,8 @@ bandwidth.
 <ol>
   <li><p>Let <var>selectedURL</var> and <var>selectedIcon</var> be null.
 
-  <li><p>For each <var>icon</var> in <var>icons</var> run these substeps:
+  <li>
+    <p>For each <var>icon</var> in <var>icons</var> run these substeps:
 
     <ol>
       <li>
--- a/notifications.bs
+++ b/notifications.bs
@@ -202,15 +202,17 @@ these steps:
   <li><p>Let <var>baseURL</var> be the API base URL specified by the
   <a>entry settings object</a>. <span class=XXX>Or incumbent?</span>
 
-  <li><p>If <var>options</var>'s <code>icon</code> is present,
-  <a lt="url parser">parse</a> it using <var>baseURL</var>, and if that does not
-  return failure, set <var>notification</var>'s <a>icon URL</a> to the return
-  value. (Otherwise <a>icon URL</a> is not set.)
+  <li><p>Run the <a>icon URL selection steps</a>, providing <var>options</var>'s
+  <code>icons</code>, <var>options</var>'s <code>icon</code>, and
+  <var>baseURL</var>, and if that does not return null, set
+  <var>notification</var>'s <a>icon URL</a> to the return value. (Otherwise
+  <a>icon URL</a> is not set.)
 
-  <li><p>If <var>options</var>'s <code>badge</code> is present,
-  <a lt="url parser">parse</a> it using <var>baseURL</var>, and if that does not
-  return failure, set <var>notification</var>'s <a>badge URL</a> to the return
-  value. (Otherwise <a>badge URL</a> is not set.)
+  <li><p>Run the <a>icon URL selection steps</a>, providing <var>options</var>'s
+  <code>badges</code>, <var>options</var>'s <code>badge</code>, and
+  <var>baseURL</var>, and if that does not return null, set
+  <var>notification</var>'s <a>badge URL</a> to the return value. (Otherwise
+  <a>badge URL</a> is not set.)
 
   <li><p>If <var>options</var>'s <code>sound</code> is present,
   <a lt="url parser">parse</a> it using <var>baseURL</var>, and if that does not
@@ -257,10 +259,11 @@ these steps:
     <li><p>Set <var>action</var>'s <a lt="action title">title</a> to the
     <var>entry</var>'s <code>title</code>.
 
-    <li><p>If <var>entry</var>'s <code>icon</code> is present,
-    <a lt="url parser">parse</a> it using <var>baseURL</var>, and if that does
-    not return failure, set <var>action</var>'s <a>action icon URL</a> to the
-    return value. (Otherwise <a>action icon URL</a> is not set.)
+    <li><p>Run the <a>icon URL selection steps</a>, providing <var>entry</var>'s
+    <code>icons</code>, <var>entry</var>'s <code>icon</code>, and
+    <var>baseURL</var>, and if that does not return null, set
+    <var>actions</var>'s <a>action icon URL</a> to the return value. (Otherwise
+    <a>action icon URL</a> is not set.)
 
     <li><p>Append <var>action</var> to <var>notification</var>'s list of
     <a>actions</a>.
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -601,6 +604,39 @@ or vibrating the device again, unless the <a>renotify preference flag</a> is set
...
+<p>The <dfn>icon URL selection steps</dfn> for selecting a URL given a list <var>icons</var> of type
+{{NotificationIcon}}, a <var>fallbackURLString</var> of type {{USVString}}, and
+<a lt="base url">base URL</a> <var>baseURL</var> are:
+
+<ol>
+  <li><p>Let <var>selectedURL</var> and <var>selectedIcon</var> be null
+
+  <li><p>For each <var>icon</var> in <var>icons</var> run these substeps:
+
+  <ol>
```
annevk: I know we have this indentation elsewhere, but if the `<p>` is not the sole child of an `<li>` it really should go on its own line and get indented. And this `<ol>` too.

notifications.bs
```diff
@@ -601,6 +604,39 @@ or vibrating the device again, unless the <a>renotify preference flag</a> is set
...
+    substeps for this <var>icon</var>
+
+    <li><p>Let <var>parsedSizes</var> be the result of <a lt="attr link sizes">parsing</a>
+    <var>icon</var>'s <code>sizes</code>
+
+    <li><p>If <var>icon</var> is considered by the implementation to be more suitable than
+    <var>selectedIcon</var>, taking into account whether <var>icon</var>'s <code>type</code> is
+    <a lt="valid mime type">valid</a> and <a lt="resources">supported</a>, as well as the value of
+    <var>parsedSizes</var>, then set <var>selectedURL</var> to <var>iconURL</var> and set
+    <var>selectedIcon</var> to <var>icon</var>
```
annevk: Where is this paragraph from?

notifications.bs
```diff
@@ -601,6 +604,39 @@ or vibrating the device again, unless the <a>renotify preference flag</a> is set
 </ol>
 
 
+<h3 id=selecting-an-icon-url>Selecting an icon URL</h3>
+
+<p>The <dfn>icon URL selection steps</dfn> for selecting a URL given a list <var>icons</var> of type
+{{NotificationIcon}}, a <var>fallbackURLString</var> of type {{USVString}}, and
+<a lt="base url">base URL</a> <var>baseURL</var> are:
+
+<ol>
```
annevk: The steps in this list need to end in dots.

annevk:
I think it would be nicer if `icon`/`badge` simply reflected the selected URL.

annevk:
(And we'd do away with `icons`/`badges`.)

notifications.bs
mvano: I wrote it, but I'm not sure whether that's what you're asking. It's similar to bits of the manifest spec, but simplified, as WebIDL does a lot of the defining for us, and we don't need to describe how to parse JSON.

mvano:
Thanks Anne, I've cleaned it up a bit as discussed. Of note:
* Just return the selected URL, do not expose getters for the multiple inputs.
* As input and output are now more obviously asymmetrical (the missing getters on the output) I had to split NotificationAction into a NotificationActionOptions dictionary and a NotificationAction interface.
* Afaict using an interface with readonly attributes for the output has the happy side effect of not needing to explicitly freeze the action, so I deleted that wrinkle.

notifications.bs
annevk: I see, I was wondering whether HTML or manifest had a similar algorithm at some point.

notifications.bs
```diff
@@ -601,6 +604,47 @@ or vibrating the device again, unless the <a>renotify preference flag</a> is set
...
+bandwidth.
+
+<p>The <dfn>icon URL selection steps</dfn> for selecting a URL given a list <var>icons</var> of type
+{{NotificationIconOptions}}, a <var>fallbackURLString</var> of type {{USVString}}, and
+<a lt="base url">base URL</a> <var>baseURL</var> are:
+
+<ol>
+  <li><p>Let <var>selectedURL</var> and <var>selectedIcon</var> be null.
+
+  <li><p>For each <var>icon</var> in <var>icons</var> run these substeps:
```
annevk: `<p>` needs to be on a new line.

notifications.bs
```diff
@@ -601,6 +604,47 @@ or vibrating the device again, unless the <a>renotify preference flag</a> is set
...
+
+    <ol>
+      <li>
+        <p>Let <var>iconURL</var> be the result of <a lt="url parser">parsing</a> <var>icon</var>'s
+        <code>src</code> using <var>baseURL</var>. If <var>iconURL</var> is failure, abort these
+        substeps for this <var>icon</var>.
```
annevk: Now here the `<p>` does not need to be on a new line, since it's the only child.

annevk:
This looks pretty good I think. Who else should we ask for review?

mvano:
CC @beverloo @foolip @mounirlamouri @xxyzzzq in case there's something really crazy in here. If necessary we can probably align a detail or two as needed in a followup change.

annevk:
I'll aim to land this Friday (maybe ping me during the day just in case) unless someone has major concerns.

foolip:
There's definitely some similarity here with https://github.com/whatwg/mediasession/pull/129

Feedback on that PR appreciated, and is there a good place to put the common bits? Looks to me at first glace like the whole "list of images" concept that uses the `[{src:'foo'}]` syntax can and should be shared.

notifications.bs
```diff
@@ -905,19 +964,11 @@ attribute's getter must return the result of the following steps:
     <li><p>Set <var>action</var>'s {{NotificationAction/icon}} to
     <var>entry</var>'s <a>action icon URL</a>.
 
-    <!-- XXX IDL dictionaries are usually returned by value, so don't need to be
-         immutable. But FrozenArray reifies the dictionaries to mutable JS
-         objects accessed by reference, so we explicitly freeze them.
-         It would be better to do this with IDL primitives instead of JS - see
-         https://www.w3.org/Bugs/Public/show_bug.cgi?id=29004 -->
```
foolip: What's the conclusion of this? Looks like the issue was side-stepped by not exposing an `icons` member on `NotificationAction`, is that a permanent situation?

mvano: Only exposing the selected URL seems much easier to understand. I don't see a strong use case for reflecting all the exact input of the options dictionaries, so I don't expect us to add that. Note that we already didn't expose the exact input of the icon urls, we reflected the result of resolving with baseURL.

Also, by using an interface with readonly attributes, I don't think we need to freeze the object anymore, which was a strange wrinkle.

foolip: So is the resource chosen synchronously so that it can be immediately reflected through a single `icon` attribute? Doesn't that mean that the bits implementing the notification API need to know quite a bit about the UI of the platform its running on?

AFACIT, the thing that was frozen here wasn't the value returned by the attribute (which is readonly), but the `NotificationAction` objects in that array. Otherwise one can do things like `notification.actions[0].icon = 'newicon.png'` which would have no effect. Exact same problem as discussed for Media Session artwork, in other words.

mvano: Good question about synchronously reflecting the selected URL... that might require an async IPC message in Chrome. I'll have to think about that one.

The general solution to avoid having to freeze dictionaries seems to be to instead return interfaces with readonly attributes. NotificationAction now is an interface that only has readonly attributes, and assigning a new value to `icon` has no effect, intentionally. The property value does not change, and internally nothing changes. No exception is thrown.

annevk:
Closing this due to lack of activity. If there's still interest let's discuss in #28.

PR #80, opened 2016-08-05
Encourage user agents to only expose supported properties

beverloo:
Addresses #81.

***
<a href="https://whatpr.org/notifications/80.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (13a81a3)">Preview</a> | <a href="https://whatpr.org/notifications/80/41843b5...13a81a3.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (13a81a3)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -291,6 +291,14 @@ removed from the <a>list of notifications</a>.
 <p>User agents should display <a>persistent notifications</a> in a platform's
 "notification center" (if available).
 
+<p>User agents should only expose properties on the {{Notification}} object when
+the associated feature impacts the user agent's presentation or behavior of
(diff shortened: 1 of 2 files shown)


notifications.bs
```diff
@@ -291,6 +291,14 @@ removed from the <a>list of notifications</a>.
 <p>User agents should display <a>persistent notifications</a> in a platform's
 "notification center" (if available).
 
+<p>User agents should only expose properties on the {{Notification}} object when
```
annevk: I think it's better to not say "on the Notification object" as they are more likely on the prototype object and maybe elsewhere.

We should probably also mention that dictionary members are likewise not to be supported (this impacts whether valueOf is run, for instance).

PR #82, opened 2016-08-05
Add image property to notifications

### LINKED ISSUE #79: Add poster (big picture / hero image)
Android, Windows and Chrome OS all support showing a poster image on a notification.

- Android calls this `bigPicture`:
https://developer.android.com/reference/android/app/Notification.BigPictureStyle.html
https://developer.android.com/design/patterns/notifications.html#ExpandedLayouts
- Windows calls this `HeroImage`:
https://blogs.msdn.microsoft.com/tiles_and_toasts/2016/05/23/notificationsextensions-updated-for-anniversary-update-of-windows-10/#toasthero
- Chrome Apps/Extensions calls this `imageUrl`:
https://developer.chrome.com/extensions/notifications#property-NotificationOptions-imageUrl
https://developer.chrome.com/apps/richNotifications

Let's add this feature to Web Notifications:
```
partial dictionary NotificationOptions {
  USVString poster;
}
```

As usual, they will be displayed on a best-effort basis. For example, both Android and Windows only display the poster image when the notification is expanded (on Android, the most recent notification is usually expanded, depending on priority; in the Windows 10 Action Center, the most recent notification shown by each app is expanded [[source](https://blogs.msdn.microsoft.com/tiles_and_toasts/2016/06/10/whats-new-for-toast-notifications-and-action-center-in-windows-10-anniversary-update/)]). Other platforms won't show them at all.

Ideally it would be possible to [feature-detect platform support](https://groups.google.com/a/chromium.org/d/msg/blink-dev/Wn_VVO3sD2k/YWYN4oV9BQAJ) by checking if Notification.prototype contains `poster`, but that's a subject for another issue. I will note that on Android at least, [`BigPictureStyle`](https://developer.android.com/reference/android/app/Notification.BigPictureStyle.html) is mutually exclusive with [`InboxStyle`](https://developer.android.com/reference/android/app/Notification.InboxStyle.html) for displaying a list of items (#21), so if we later add lists to Web Notifications we may need some way to indicate that they are mutually exclusive and/or allow web developers to indicate which property is more important, but we can cross that bridge later.

johnmellor:
Resolves #79


notifications.bs
```diff
@@ -93,23 +96,39 @@ that the end user should not be able to easily clear the
 <a lt="concept notification">notification</a> <span class=note>Only makes sense
 for <a>persistent notifications</a>.
 
-<p>A <a lt="concept notification">notification</a> <em>can</em> have an
-associated <dfn>icon URL</dfn>, <dfn>icon resource</dfn>, <dfn>sound URL</dfn>,
-<dfn>sound resource</dfn>, <dfn>vibration pattern</dfn>, and
-<dfn>service worker registration</dfn>.
+<p>A <a lt="concept notification">notification</a> <em>can</em> have up to three
+associated graphics: an <dfn>image URL</dfn>, <dfn>icon URL</dfn>, and
```
annevk: up to three -> these

annevk:
This looks reasonable to me now.

annevk:
@inscriber do you want to take a look? @overholt said you're sorta in charge of notifications.

beverloo:
Friendly ping. Is there anything we should do to move this along?

annevk:
Sorry about the slow follow up. Should be all good now.

PR #85, opened 2016-11-04
Remove sticky and screen off features

annevk:
They are not implemented and there’s no interest per Peter Beverloo. We
can reinstate these if that changes.


annevk:
@beverloo okay to merge this?

beverloo:
Looks good to me - thanks for doing this!

PR #88, opened 2016-12-07
Fixing ambiguous use of response

jakearchibald:
Fixes minor bikeshed warning.


annevk:
Thanks, I ended up fixing this myself since I found a couple more issues and I'd rather use `for` than any kind of defaulting.

PR #89, opened 2016-12-07
Add note about intent of fetching

jakearchibald:
There's a link as an HTML comment, but feels like it should be visible when reading the spec.

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -375,90 +375,116 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
 <a>notification</a> <var>notification</var> are:
 
 <ol>
-  <!-- XXX https://www.w3.org/Bugs/Public/show_bug.cgi?id=24055 -->
-  <li><p>If the notification platform supports images, <a for=/>fetch</a>
+ <!-- XXX https://www.w3.org/Bugs/Public/show_bug.cgi?id=24055 -->
+ <li>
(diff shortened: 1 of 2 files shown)


notifications.bs
```diff
@@ -379,6 +379,11 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
   <li><p>If the notification platform supports images, <a for=/>fetch</a>
   <var>notification</var>'s <a>image URL</a>, if <a>image URL</a> is set.
 
+  Note: The intent is to fetch this resource similar to an
+  <a href="https://html.spec.whatwg.org/multipage/embedded-content.html#update-the-image-data">
+    <code>&lt;img&gt;</code>
+  </a>, but this <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=24055">needs abstracting</a>.
```
annevk: Can you make these `<p class=note>` instead of `Note:`?

The other `<p>` inside the `<li>` will also need to come on its own line since the `<li>` now contains multiple children and indented with one more space.

The `<a>` and its following comma should all be on a single line with no spaces between the brackets of `<a>` and `<code>`.

The same formatting applies to the other changes made here.

Thanks!

jakearchibald:
@annevk I fixed the indenting in that section in a separate commit - it's up to you if you want to keep it.

PR #111, opened 2017-10-20
Add a warning about the reliability of the notificationclose event

### LINKED ISSUE #109: Required steps for closing a notification are incompatible with some platforms
On Apple platforms when the user dismisses an app's notification without activating it, the app is never woken up to be told that. No client-side app code is run in this case; the app is only woken up upon activation of the notification.

Could the wording of [§2.8 Closing a notification](https://notifications.spec.whatwg.org/#closing-a-notification) be loosened to handle such cases?

beverloo:
Not all platforms dispatch an event to the applications (e.g. the user
agent) when a notification is dismissed, or only do so in a subset of
cases. This means that the `notificationclose` event is a best-effort
event. It adds significant value to the developer, as dismissal is a
powerful (yet complicated) signal, so user agents should still make an
effort to support it where they can.

Fixes #109

***
<a href="https://whatpr.org/notifications/111.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (8e39f86)">Preview</a> | <a href="https://whatpr.org/notifications/111/439a225...8e39f86.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (8e39f86)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -582,6 +582,10 @@ must be run.
   <li><p>Remove <var>notification</var> from the <a>list of notifications</a>.
 </ol>
 
+<p class="warning">Because not all notification centers reliably dispatch an event when a
+notification is closed, the <code>notificationclose</code> event will be delivered on a best-effort
+basis. User agents should aim to support the event whenever possible.
+
 
 <h3 id=displaying-notification>Displaying notifications</h3>
 


beverloo:
PTAL @annevk and @hober

I decided to not amend the algorithm as the user agent would not know about either closing of the notification, or about whether it was user triggered, when this situation arises. Does that sound reasonable?

hober:
The proposed change leaves the RFC 2119 MUST in place, though. I don't think this change is sufficient.

beverloo:
Just changing the `MUST` to `SHOULD` imposes a significant change to the _list of notifications_ concept. It would either be stale as deletion becomes optional (how will `getNotifications()` behave?) or would become a shared responsibility between the user agent and the underlying platform notification center. The second option means that the user agent _will_ be aware of closed notifications, just not in a timely manner.

Instead, maybe we should somehow rephrase the event dispatching section (2.8.2.) to be limited to situations where the UA does know in a timely manner?

annevk:
Maybe something like:

> When a notification is closed, either by the underlying notification platform or by the user, and the underlying notification platform exposes notification closure, the close steps for it must be run.

PR #121, opened 2018-02-20
Restrict use of Notification.requestPermission to secure contexts.

dominickng:
***
<a href="https://whatpr.org/notifications/121.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (fcf5c38)">Preview</a> | <a href="https://whatpr.org/notifications/121/4a296ed...fcf5c38.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (fcf5c38)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -1210,7 +1210,7 @@ urlPrefix: https://w3c.github.io/vibration/
 urlPrefix: https://tc39.github.io/ecma262/#sec-object.; type: dfn
     text: freeze
 urlPrefix: https://w3c.github.io/webappsec-secure-contexts/; type: dfn
-    text: secure context
+    text: secure context; url: #secure-contexts
 </pre>
 
 <pre class="biblio">
--- a/notifications.bs
+++ b/notifications.bs
@@ -775,8 +775,12 @@ method, when invoked, must run these steps:
    <a>entry settings object</a>'s <a for="environment settings object">origin</a>.
    <!-- XXX context object's global object's environment settings object's origin -->
 
-   <li><p>If <var>permission</var> is "<code>default</code>", ask the user whether showing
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -775,8 +775,12 @@ method, when invoked, must run these steps:
    <a>entry settings object</a>'s <a for="environment settings object">origin</a>.
    <!-- XXX context object's global object's environment settings object's origin -->
 
-   <li><p>If <var>permission</var> is "<code>default</code>", ask the user whether showing
-   notifications for the <a>entry settings object</a>'s
+   <li><p>If the <a>entry settings object</a> is not a <a>secure context</a>, then set
+   <a>permission</a> for the <a>entry settings object</a>'s
+   <a for="environment settings object">origin</a> to <var>denied</var>.
```
domenic: Please use current, not entry; entry is deprecated.

dominickng:
I'd also like to augment the secure context restriction with a top-level context restriction, i.e. Notification.requestPermission returns "denied" for non-secure and non-top-level contexts. Chrome shipped both of these restrictions in tandem.

I can pull the top-level discussion out to a separate PR (and pull it out of the corresponding WPT at https://github.com/w3c/web-platform-tests/pull/6596) if so desired, but it would be nice to combine them.

annevk:
If `Notification.permission` is to return "denied" by default, we also need to update the algorithm of its getter. Otherwise it would only return "denied" if you first invoked `requestPermission()`. (We should probably test that too.)

Do you mean cross-origin non-top-level contexts or non-top-level contexts in general? We should probably do that as a separate change and couple it with Feature Policy somehow.

annevk:
This got fixed by #176.

PR #123, opened 2018-02-27
Editorial: adopt more Infra terminology

annevk:
***
<a href="https://whatpr.org/notifications/123.html" title="Last updated on Feb 27, 2018, 3:12 PM GMT (e2c930a)">Preview</a> | <a href="https://whatpr.org/notifications/123/1961618...e2c930a.html" title="Last updated on Feb 27, 2018, 3:12 PM GMT (e2c930a)">Diff</a>

PR #124, opened 2018-02-27
getNotifications() needs to do work on the "main thread" 

annevk:
Once #123 lands this can be rebased and landed.

***
<a href="https://whatpr.org/notifications/124.html" title="Last updated on Apr 22, 2021, 12:54 PM UTC (8df2f2f)">Preview</a> | <a href="https://whatpr.org/notifications/124/a3ddf9f...8df2f2f.html" title="Last updated on Apr 22, 2021, 12:54 PM UTC (8df2f2f)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -9,6 +9,14 @@ Translation: ja https://triple-underscore.github.io/notifications-ja.html
 Translation: zh-cn https://w3c-html-ig-zh.github.io/notifications/whatwg/
 </pre>
 
+<!-- Remove this once https://github.com/whatwg/html/pull/6360 is resolved. -->
+<pre class=anchors>
+urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html#; spec: html;
+    type: dfn;
+        text: DOM manipulation task source; url: dom-manipulation-task-source
+</pre>
+
+
 <h2 id=terminology>Terminology</h2>
 
 <p>This specification depends on the Infra Standard. [[!INFRA]]
--- a/notifications.bs
+++ b/notifications.bs
@@ -721,36 +721,41 @@ permissions are like synchronous IO, a bad idea.
 
 <p>The static
 <dfn method for=Notification><code>requestPermission(<var>deprecatedCallback</var>)</code></dfn>
-method, when invoked, must run these steps:
+method steps are:
 
 <ol>
- <li><p>Let <var>promise</var> be a new promise.
+ <li><p>Let <var>global</var> be <a>this</a>'s <a>relevant global object</a>.
+
+ <li><p>Let <var>origin</var> be <a>this</a>'s <a>relevant settings object</a>'s
+ <a for="environment settings object">origin</a>.
+ <!-- Arguably this should snapshot origin somehow so that changes to it do not impact the
+      in parallel part. -->
+
+ <li><p>Let <var>promise</var> be <a for=/>a new promise</a> in <var>global</var>'s
+ <a for="global object">Realm</a>.
 
  <li>
   <p>Run these steps <a>in parallel</a>:
 
   <ol>
-   <li><p>Let <var>permission</var> be <a>permission</a> for
-   <a>entry settings object</a>'s <a for="environment settings object">origin</a>.
-   <!-- XXX context object's global object's environment settings object's origin -->
+   <li><p>Let <var>permission</var> be <a>permission</a> for <var>origin</var>.
 
-   <li><p>If <var>permission</var> is "<code>default</code>", ask the user whether showing
-   notifications for the <a>entry settings object</a>'s
-   <a for="environment settings object">origin</a> is acceptable. If it is, set
-   <var>permission</var> to "<code>granted</code>", and "<code>denied</code>" otherwise.
+   <li><p>If <var>permission</var> is "<code>default</code>", then ask the user whether showing
+   notifications for <var>origin</var> is acceptable. If it is, then set <var>permission</var> to
+   "<code>granted</code>"; otherwise "<code>denied</code>".
 
    <li>
-    <p><a>Queue a task</a> to run these steps:
+    <p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given <var>global</var>
+    to run these steps:
 
     <ol>
-     <li><p>Set <a>permission</a> for the <a>entry settings object</a>'s
-     <a for="environment settings object">origin</a> to <var>permission</var>.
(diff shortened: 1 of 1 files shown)


annevk:
@bzbarsky since you raised the issue, care to review? You can use the Preview and Diff links in OP for a rendered version.

bzbarsky:
I don't see anything that allows resolving a Promise with an infra list.  We should probably add something in either infra spec or promises guide.  Probably it would be enough to say that you can resolve a Promise with a WebIDL `sequence<T>` and that will apply <https://heycam.github.io/webidl/#sequence-to-es>.  Given the bit at https://heycam.github.io/webidl/#idl-sequence about single-type lists being implicitly considered sequences that would work.

Do we need to specify a task source for the task?

Do we need to specify the global used for the Notification objects?

annevk:
@bzbarsky I think IDL needs to define "resolve" and "reject" for promises, possibly HTML as some folks want "queue a task and resolve" semantics.

I never quite know what task source to use or what principles we use for them. I was hoping to at some point formalize this a bit better in HTML and then go through all callers to make sure they make sense. (E.g., I think we should make "queue a task" just take a source as argument.) It seems this could be its own task source, if we use the same one for all notification-related tasks.

I think using the current Realm would be fine (and I believe the plan was for that to be the default to be defined in IDL), but maybe @domenic has thoughts.

bzbarsky:
https://www.w3.org/2001/tag/doc/promises-guide defines resolve and reject, but needs ES values at the moment.

> I never quite know what task source to use or what principles we use for them.

Me neither, really...  Basically, things that should be ordered wrt each other should use the same task source.  Past that, who knows.  :(

> I think using the current Realm would be fine

There is no current Realm, because you're in an async section.

annevk:
@bzbarsky at the moment those steps run you're no longer "in parallel", but it does seem undefined what the realm is when a task runs. I guess "queue a task" needs to be both explicit about the realm and the event source...

annevk:
(As for promises, I think they should work the same as returning from a method and therefore IDL needs to handle the conversion from a list to a JavaScript type.)

bzbarsky:
> but it does seem undefined what the realm is when a task runs

In particular, when a task runs there is no current Realm.  If you need a realm in there, you need to save it before queuing the task or going parallel.

I agree on how promises should work.  We just need to adjust promises-guide to say so.  We'd need to double-check that there is never ambiguity, given an infra type of some sort, about what the JS reflection looks like (that is, that return types in IDL really are just documentation, not behavior-changing).

rakuco:
Is there anything blocking this PR from being merged? I was going to file an issue and noticed this was already being worked on. Note that showNotification() is also resolving/rejecting a promise in parallel.

annevk:
The outstanding comments on realms and task source, mostly, but I think I have an idea as to what the answers should be.

annevk:
I don't understand why DOM manipulation task source doesn't link. It says it's not a dfn, but HTML exports it and hasn't set any attributes making it not a dfn, as far as I can tell. @tabatkins thoughts?

tabatkins:
I presume it's because its ID starts with "dom-". HTML gets heuristics applied to it because it doesn't follow the dfn contract faithfully, and one (usually very reliable) heuristic is that "dom-*" IDs are referring to WebIDL constructs. It tries to match them against IDL blocks to figure out if it's a method or attribute or whatever, and failing that, marks them as an interface.

annevk:
Would it stop applying the heuristic if I set dfn-type?

tabatkins:
Yup, the heuristics fire to determine what the type is; if you specify the type it's fine.

notifications.bs
```diff
@@ -721,36 +721,41 @@ permissions are like synchronous IO, a bad idea.
...
- <li><p>Let <var>promise</var> be a new promise.
+ <li><p>Let <var>global</var> be <a>this</a>'s <a>relevant global object</a>.
+
+ <li><p>Let <var>origin</var> be <a>this</a>'s <a>relevant settings object</a>'s
+ <a for="environment settings object">origin</a>.
+ <!-- Arguably this should snapshot origin somehow so that changes to it do not impact the
+      in parallel part. -->
+
+ <li><p>Let <var>promise</var> be <a for=/>a new promise</a> in <var>global</var>'s
+ <a for="global object">Realm</a>.
```
domenic, line 735: This is fine but you could also do "a new promise in this's relevant Realm"

notifications.bs
```diff
@@ -1002,69 +1007,90 @@ partial interface ServiceWorkerGlobalScope {
...
-    <li><p>If <a>permission</a> for <var>notification</var>'s <a for=notification>origin</a> is not
-    "<code>granted</code>", then reject <var>promise</var> with a <code>TypeError</code> exception,
-    and abort these steps.
+   <li><p>If <a>permission</a> for <var>notification</var>'s <a for=notification>origin</a> is not
+   "<code>granted</code>", then <a>queue a global task</a> on the
+   <a>DOM manipulation task source</a> given <var>global</var> to <a for=/>reject</a>
+   <var>promise</var> with a {{TypeError}}, and abort these steps.
 
-    <li><p>Run the <a>fetch steps</a> for <var>notification</var>.
+   <li><p>Run the <a>fetch steps</a> for <var>notification</var>.
```
domenic, line 1040: Probably a separate issue but the fetching also needs some modernization on top of the various processResponse/useParallelQueue stuff..

notifications.bs
```diff
@@ -1002,69 +1007,90 @@ partial interface ServiceWorkerGlobalScope {
...
+      in parallel part. -->
+
+ <li><p>Let <var>promise</var> be <a for=/>a new promise</a> in <var>realm</var>.
+
+ <li>
+  <p>Run these steps <a>in parallel</a>:
 
   <ol>
-    <li><p>Let <var>tag</var> be <var>filter</var>'s <code>tag</code>.
+   <li><p>Let <var>tag</var> be <var>filter</var>'s <code>tag</code>.
```
domenic, line 1071: This could use `<var>filter</var>["{{GetNotificationOptions/tag}}"]`

domenic (approved):
I'd love to see this landed; I was working on https://github.com/jeremyroman/alternate-loading-modes/pull/57 and noticed the algorithms I was monkeypatching would benefit from a bit of modernization :)

annevk:
Thanks for the review. I pushed a fixup commit. This is still blocked on https://github.com/whatwg/html/pull/6360 unfortunately. I pinged that again.

domenic:
I mean, it's not really blocked; you could override with link-defaults or anchors?

PR #125, opened 2018-02-27
Use SetIntegrityLevel instead of Object.freeze()

### LINKED ISSUE #74: "Call Object.freeze on action"
https://notifications.spec.whatwg.org/#object-members

> Call Object\.freeze on action, to prevent accidental mutation by scripts\. 

This should not call the author-overrideable Object.freeze. Instead, call [SetIntegrityLevel](https://tc39.github.io/ecma262/#sec-setintegritylevel)(action, "frozen").

annevk:
Verbose, but clear and correct.

Fixes #74.

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -909,13 +909,13 @@ return the <a>notification</a>'s <a for=notification>require interaction prefere
 result of the following steps:
 
 <ol>
-  <li><p>Let <var>frozenActions</var> be an empty list of type {{NotificationAction}}.
+  <li><p>Let <var>frozenActions</var> be a new <a for=/>list</a>.
 
   <li><p>For each <var>entry</var> in the <a>notification</a>'s list of
   <a for=notification>actions</a>, perform the following steps:
 
   <ol>
-    <li><p>Let <var>action</var> be a new {{NotificationAction}}.
+    <li><p>Let <var>action</var> be a new {{NotificationAction}} <a for=/>dictionary</a>.
 
     <li><p>Set <var>action</var>'s {{NotificationAction/action}} to
     <var>entry</var>'s <a for=action>name</a>.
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -926,15 +926,15 @@ result of the following steps:
...
-         It would be better to do this with IDL primitives instead of JS - see
-         https://www.w3.org/Bugs/Public/show_bug.cgi?id=29004 -->
-    <li><p>Call <a lt="freeze">Object.freeze</a> on <var>action</var>, to
-    prevent accidental mutation by scripts.
+    <li><p>Let <var>jsAction</var> be <var>action</var>, <a>converted to an ECMAScript value</a>.
 
-    <li><p>Append <var>action</var> to <var>frozenActions</var>.
+    <li><p>Perform <a abstract-op>SetIntegrityLevel</a>(<var>jsAction</var>, "frozen").
+
+    <li><p>Let <var>idlAction</var> be <var>jsAction</var>, <a>converted to an IDL value</a>.
```
domenic: This isn't correct, at least as written. The conversion, presumably to an IDL `NotificationAction` type (although that isn't stated), will re-perform the Get()s of all the properties, which can be observable if Object.prototype is modified. Then we'll convert back to JS in the binding layer.

I don't think there's any way to do this rigorously as long as your return type is `FrozenArray<NotificationAction>` (instead of `FrozenArray<object>`). The conversion between JS objects and dictionaries is observable and side effecting.

annevk: I cannot use `FrozenArray` then either since it's constructor takes an IDL value. I'm not sure I want to inline all of that. That would be rather bad and I'd likely make mistakes or it would result in mismatches down the line.

There is some mismatch here though with implementations I think where JS values are backed by IDL values and you can manipulate either. That's not conversion, but just grabbing the corresponding value, but IDL doesn't formalize that.

PR #126, opened 2018-02-28
Merge replacing and displaying into showing

annevk:
This gives a more consistent story for event dispatching. The former setup was especially buggy if native replacement is not supported.

This also reduces duplication (such as waiting for fetches).

And it queues a task for the show and close events.

***
<a href="https://whatpr.org/notifications/126.html" title="Last updated on Feb 28, 2018, 6:28 PM GMT (46a1553)">Preview</a> | <a href="https://whatpr.org/notifications/126/459bf35...46a1553.html" title="Last updated on Feb 28, 2018, 6:28 PM GMT (46a1553)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -70,7 +70,7 @@ run.
 
 <p>A <a>notification</a> has an associated
 <dfn for=notification id=silent-preference-flag>silent preference flag</dfn> which is initially
-unset. When set indicates that no sounds or vibrations should be made.
+unset. When set indicates that no vibrations should be made.
 
 <p>A <a>notification</a> has an associated
 <dfn for=notification id=require-interaction-preference-flag>require interaction preference flag</dfn>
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -480,15 +484,59 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
+ <a for=notification>actions</a> (if any).
+
+ <li><p>Let <var>shown</var> be false.
+
+ <li>
+  <p>If <var>notification</var>'s <a for=notification>old notification</a> is non-null, then:
+
+  <ol>
+   <li><p><a>Handle close events</a> with <var>notification</var>'s
+   <a for=notification>old notification</a>.
```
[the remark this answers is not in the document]
annevk: Good catch. I think that once #127 lands I could rebase and rewrite this bit to query the old notification during showing. That should remove this issue and leave less state on notifications.

The only thing that remains to fully remove all race conditions is to use a "parallel queue" to ensure that the "in parallel" steps don't race with each other. I need to file a follow-up on that.

annevk:
I rebased this on top of the sound removal and thereby addressed the race issue (I hope). I filed #128 as a follow-up to formalize even more.

notifications.bs
```diff
@@ -451,15 +445,62 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
 <p>The <dfn>show steps</dfn> for a given
 <a>notification</a> <var>notification</var> are:
+<!-- These steps are invoked from "in parallel" steps -->
 
 <ol>
- <li><p>If <var>notification</var> is <a>replaceable</a>, then run the <a>replace steps</a> for that
- <a>notification</a> and <var>notification</var>.
+ <li><p>Wait for any <a for=/ lt=fetch>fetches</a> to complete and <var>notification</var>'s
+ <a for=notification>image resource</a> <a for=notification>icon resource</a>,
+ <a for=notification>badge resource</a>, and <a for=notification>sound resource</a> to be set
```
beverloo: nit: no more sound resource

notifications.bs
```diff
@@ -70,7 +70,7 @@ run.
 
 <p>A <a>notification</a> has an associated
 <dfn for=notification id=silent-preference-flag>silent preference flag</dfn> which is initially
-unset. When set indicates that no sounds or vibrations should be made.
```
beverloo: I kept this in because the user agent could still suppress any default sounds the platform's notification center may make, which feels appropriate for a property called `silent`.

annevk: Ah, good call.

PR #127, opened 2018-02-28
Remove support for sounds from notifications

beverloo:
There are no implementations for this property, and the native
notification centers of various platforms make it hard to support this
properly.

  - Windows Action Center toasts have an <audio> element supports
    built-in sounds, as well as sounds included in an application's
    package, but not freeform path and/or URL support.
  - NSUserNotification (Mac OS and iOS) have a `soundName` property that
    will be resolved against files in various local system directories.
  - Android strongly prefers usage of the RingtoneManager for sounds.
  - Based on Chrome for Linux statistics, the majority of Chrome users (80%+)
    report that sounds are not supported in the system's notification server.

Even though it could be supported on certain platforms, it's currently
underspecified: which codecs should be supported and what is the maximum
duration of the sound (if any), making it very unpredictable at best.

We can revisit supporting sounds if there is interest from
implementations.

***
<a href="https://whatpr.org/notifications/127.html" title="Last updated on Feb 28, 2018, 1:45 PM GMT (8a7bd8a)">Preview</a> | <a href="https://whatpr.org/notifications/127/daf4b50...8a7bd8a.html" title="Last updated on Feb 28, 2018, 1:45 PM GMT (8a7bd8a)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -99,14 +99,12 @@ not enough space to display the <a>notification</a> itself. It <em>may</em> also
 the <a>notification</a>, but then it should have less visual priority than the
 <a for=notification>image resource</a> and <a for=notification>icon resource</a>.
 
-<p>A <a>notification</a> <em>can</em> have a <dfn for=notification id=sound-url>sound URL</dfn>,
-<dfn for=notification id=sound-resource>sound resource</dfn>,
+<p>A <a>notification</a> <em>can</em> have a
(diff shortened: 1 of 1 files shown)


beverloo:
What do you think, @annevk? I'll send a PR to update WPT if this is reasonable.

annevk:
Seems good to me.

beverloo:
WPT PR: https://github.com/w3c/web-platform-tests/pull/9713

PR #132, opened 2018-05-22
Add support for inline replies

anitawoodruff:
This change adds support for notifications that accept text input within their UI, aka 'inline replies'.

Explainer: https://github.com/anitawoodruff/inline-notification-replies

Requesting TAG review shortly, will link to it below.

***
<a href="https://whatpr.org/notifications/132.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (713eafb)">Preview</a> | <a href="https://whatpr.org/notifications/132/e9407eb...713eafb.html" title="Last updated on Jan 15, 2021, 7:46 AM UTC (713eafb)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -534,7 +534,8 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
 
 <p class="note">User agents are required to attribute notifications to the origin they
 are associated with. This reduces the possibility of phishing attacks where a notification has been
-designed to look like it originates from another origin.
+designed to look like it originates from another origin. The user should be aware that data entered
+into the notification will be made available to the attributed origin.
 
 <h3 id=activating-a-notification>Activating a notification</h3>
 


anitawoodruff:
TAG review now requested at: https://github.com/w3ctag/design-reviews/issues/284

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
-zero or more <dfn for=notification lt=action id=actions>actions</dfn>. Each
-<a for=notification>action</a> has an associated
-<dfn for=action id=action-title>title</dfn> and <dfn for=action id=action-name>name</dfn> and
-<em>can</em> have an associated <dfn for=action>icon URL</dfn> and
-<dfn for=action>icon resource</dfn>. Users may activate actions, as alternatives to
-activating the notification itself. The user agent must determine the
-<dfn>maximum number of actions</dfn> supported, within the constraints of the
-notification platform.
+<p>A <a>notification</a> has <dfn for=notification lt=action id=actions>actions</dfn>, 
+a <a for=/>list</a> of <a>notification actions</a>, initially empty.
```
annevk: Can we change this to something like "A notification has an action list (a list of actions). It is initially empty." And then use the `<dfn>` instead of the the `<dfn>` for "notification action" below since "action" always meant to refer to the individual item and not the associated list.

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+
+<p>A <dfn id=notification-action>notification action</dfn> defines a way the user can interact with
+a notification, as an alternative to activing the notification itself.
+
+<div dfn-for="notification action">
+  <p>A <a>notification action</a> has a <dfn id=notification-action-type>type</dfn>, that is
```
annevk: Please use one space for indentation.

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+  "<code>button</code>" or "<code>text</code>".
+  
+  <p class=note>Actions of type "<code>button</code>" can only be activated, whereas actions of type
+  "<code>text</code>" allow the user to input text during activation.
+
+  <p>A <a>notification action</a> has a <dfn id=notification-action-title>title</dfn> (a DOMString).
```
annevk: Please don't change the ID. That'll affect folks linking to this.

Thanks for adding a type, but let's make that "a string".

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+
+  <p>A <a>notification action</a> has a <dfn id=notification-action-name>name</dfn> (a DOMString).
```
annevk: Same comments as above and they apply below too.

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+
+<p>The user agent must determine the <dfn>maximum number of actions</dfn> supported, within the
+constraints of the notification platform.
 
 <p class=note>Since display of actions is platform-dependent, developers are
 encouraged to make sure that any action a user can invoke from a notification is
-also available within the web application.
+also available within the web application. The ability to reply inline to
+notifications during activation is also platform-dependent, so developers are encouraged to
+handle the case where a text action was activated but the reply was null (e.g.
```
annevk: Follow e.g. with a comma.

notifications.bs
```diff
@@ -227,14 +250,22 @@ these steps:
   <li><p>If <var>options</var>'s <code>requireInteraction</code> is true, set
   <var>notification</var>'s <a for=notification>require interaction preference flag</a>.
 
-  <li><p>Set <var>notification</var>'s list of <a>actions</a> to an empty list,
-  then for each <var>entry</var> in <var>options</var>'s <code>actions</code>,
-  up to the <a>maximum number of actions</a> supported (skip any excess
-  entries), perform the following steps:
+  <li>For each <var>entry</var> in <var>options</var>'s <code>actions</code>:
```
annevk: `<li><p>` throughout.

notifications.bs
```diff
@@ -227,14 +250,22 @@ these steps:
...
 
   <ol>
+    <li>If the user agent does not support additional actions, break.
```
annevk: then break*

notifications.bs
```diff
@@ -227,14 +250,22 @@ these steps:
...
+
     <li><p>Let <var>action</var> be a new <a lt="actions">action</a>.
 
+    <li><p>Set <var>action</var>'s <a for=action>type</a> to the <var>entry</var>'s
+    <code>type</code> property.
+
+    <li>If the user agent does not support additional actions of this
+    <a for="notification action">type</a>, continue.
```
annevk: then continue*

notifications.bs
annevk: Also, no need to set an ID for new `<dfn>`s.

notifications.bs
```diff
@@ -517,11 +552,18 @@ specified) run these steps:
...
-   <li><p>Let <var>action</var> be the empty string.
+   <li><p>Let <var>action</var> be an empty string.
+
+   <li><p>Let <var>reply</var> be null.
 
    <li><p>If one of <var>notification</var>'s <a for=notification>actions</a> was activated by the
    user, then set <var>action</var> to that <a for=notification>action</a>'s <a for=action>name</a>.
 
+   <li><p>If one of <var>notification</var>'s <a for=notification>actions</a>
+   with type <i>'text'</i> was activated by the user, and the opportunity to
```
annevk: This should be "`<code>text</code>`"?

notifications.bs
```diff
@@ -1087,8 +1143,10 @@ the same underlying <a>notification</a> of {{Notification}} objects already in e
 given <var>notification</var> and <var>action</var>,
 <a>fire an event</a> named <var>e</var>, using {{NotificationEvent}}, with the
 {{NotificationEvent/notification}} attribute initialized to a new
-{{Notification}} object representing <var>notification</var> and the
-{{NotificationEvent/action}} attribute initialized to <var>action</var>.
+{{Notification}} object representing <var>notification</var>, the
+{{NotificationEvent/action}} attribute initialized to <var>action</var>, and the
+{{NotificationEvent/reply}} attribute initialized to <var>reply</var>.
```
annevk: Where does the reply variable come from?

anitawoodruff: Ah good point, guess I can just change the beginning of this sentence to 'given notification, action and reply' which should cover it, yes?

annevk: Well, you'll need to update the callers of this algorithm too, then.

notifications.bs
anitawoodruff: I'm a little confused here about where to put the `<dfn>` tags from your comment, can you clarify?

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done.

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Done

notifications.bs
anitawoodruff: Ah yes, that makes sense.

Have updated Step 1.5 of Section 2.7 'Activating a Notification' to 'given notification, action and reply' also. I think this now covers it.

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
-<p>A <a>notification</a> has an associated list of
-zero or more <dfn for=notification lt=action id=actions>actions</dfn>. Each
-<a for=notification>action</a> has an associated
-<dfn for=action id=action-title>title</dfn> and <dfn for=action id=action-name>name</dfn> and
-<em>can</em> have an associated <dfn for=action>icon URL</dfn> and
-<dfn for=action>icon resource</dfn>. Users may activate actions, as alternatives to
-activating the notification itself. The user agent must determine the
-<dfn>maximum number of actions</dfn> supported, within the constraints of the
-notification platform.
+<p>A <a>notification</a> has an action list (a <a for=/>list</a> of
```
annevk: "action list" needs to be a `<dfn>` (and some places that currently refer to "actions" need to be updated accordingly)

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+<dfn for=notification lt=action id=actions>actions</dfn>). It is initially empty.
```
annevk: This should become a reference `<a for=notification>action</a>` and then this `<dfn>` can replace the `<dfn>` below so we just have "action" and not "notification action".

anitawoodruff:
Go for it, any time is fine - from my perspective the changes are done aside from your comments :)

notifications.bs
annevk: There's two callers of this algorithm, so I don't think that quite covers it. Also, if both are going to set it to null, does it need to be a variable for this algorithm?

annevk:
I addressed some of the formatting issues. There's still at least one outstanding issue though. Review from @beverloo would also be good.

What we need beyond this:
* Implementer interest beyond Chrome.
* Bugs filed against browsers as per https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests

notifications.bs
anitawoodruff: Step 1.4 of Section 2.7 ('Activating a notification') sets 'reply' to the text entered by the user during activation, so it shouldn't be null in that case.

I assume you are talking about Section 2.8 ('Closing a notification') as the other caller of this algorithm. That caller does not pass anything for 'action' either right now, so I guess this needs fixing too?

annevk: Ah yeah :/ I guess action needs to be the empty string there; @beverloo can confirm.

notifications.bs
```diff
@@ -108,19 +108,42 @@ the <a>notification</a>, but then it should have less visual priority than the
...
+ <p>An <a for=notification>action</a> has a <dfn id=action-name>name</dfn> (a string).
+
+ <p>An <a for=notification>action</a> has an <dfn>icon URL</dfn> (null or a <a>URL</a>). Unless
+ stated otherwise, it is null.
+
+ <p>An <a for=notification>action</a> has an <dfn>icon resource</dfn> (null or an image). Unless
+ stated otherwise, it is null.
+
+ <p>An <a for=notification>action</a> has a <dfn>placeholder</dfn> (a string). Unless stated
+ otherwise, it is an empty string.
```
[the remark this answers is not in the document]
anitawoodruff: Done.

notifications.bs
```diff
@@ -227,27 +250,36 @@ these steps:
...
+   <li><p>Let <var>action</var> be a new <a for=notification>action</a>.
+
+   <li><p>Set <var>action</var>'s <a for=action>type</a> to <var>entry</var>'s <code>type</code>
+   member.
 
-    <li><p>Set <var>action</var>'s <a for=action>name</a> to the <var>entry</var>'s
-    <code>action</code>.
+   <li>
+    <p>If the user agent does not support additional actions of this <a for=action>type</a>, then
+    continue.
```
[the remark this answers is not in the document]
anitawoodruff: ok; removed the stuff here and am adding the following to section 2.6 (Showing a notification), as a note:

"If the platform does not support entering text within a
 notification, implementers are expected to treat text actions as regular buttons and update their
 type to <code>button</code> accordingly."

How does that sound?

notifications.bs
anitawoodruff: Hm yeah that would involve adding a whole new interface to the ServiceWorker API, no?

I'm up for doing this but won't this involve quite a lot of work updating idls and tests etc? Could it be done as a followup in a separate change?

notifications.bs
```diff
@@ -227,27 +250,36 @@ these steps:
...
-  <li><p>Set <var>notification</var>'s list of <a>actions</a> to an empty list,
-  then for each <var>entry</var> in <var>options</var>'s <code>actions</code>,
-  up to the <a>maximum number of actions</a> supported (skip any excess
-  entries), perform the following steps:
+ <li>
+  <p>For each <var>entry</var> in <var>options</var>'s <code>actions</code>:
 
   <ol>
-    <li><p>Let <var>action</var> be a new <a lt="actions">action</a>.
+   <li><p>If the user agent does not support additional actions, then break.
```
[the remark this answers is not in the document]
anitawoodruff: Done. (And sorry for not 'starting a review' for these comments!)

anitawoodruff:
Filed the following implementation bugs against the other browsers:

- Edge: https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34442866-implement-inline-replies-for-web-notifications 

- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1466499

- Safari: https://bugs.webkit.org/show_bug.cgi?id=186265

anitawoodruff:
(Chrome bug is at https://crbug.com/599859 )

notifications.bs
anitawoodruff: Agreed with beverloo@ offline that changing the event hierarchy is out of scope for this change.

Have now updated the 'Closing a notification' algorithm as suggested to pass the empty string & null.

anitawoodruff:
@aliams, @hober, @annevk - Please can you comment on whether your respective browsers would be interested in implementing this feature.

Since the native notification centers on Mac, Android, iOS and Windows now all provide this functionality, we think it would make a great addition to the spec.

See [the explainer](https://github.com/anitawoodruff/inline-notification-replies) for more details.

annevk:
@anitawoodruff the main input I got at https://github.com/mozilla/standards-positions/issues/91#issuecomment-394271102 from @martinthomson was that it was unclear how to deploy this in a backwards compatible fashion. That is, how can a website use this feature in a way that doesn't make things weird in browsers that don't support it?

annevk:
FWIW, @jakearchibald offered to help with rebasing this. I think the main thing lacking here is explicit interest from one more browser. And perhaps some non-IDL tests.

sorin-davidoi:
Maybe it would be useful to expand the API to allow the reply to be prefilled with some value?

Use-case:
- User `A` sends a message to user `B`: `@b how are you?`
- User `B` wants to reply to the message.

It is common in some platforms (such as Mastodon and Twitter) for `B`'s reply to start with `@a`, and when replying to such a message the input field is prepopulated with `@a`.

This is not possible to achieve with the current API.

karlhorky:
Just as a note from us all the way over here in userland, I was just wondering about this particular feature today (whether it was possible with the Notification API):

https://twitter.com/karlhorky/status/1167778767542128643

**Use Case**: It came up as part of a discussion about whether a Slack PWA could hold its own against an Electron app: 

https://twitter.com/karlhorky/status/1167777816466866177

Thanks for listening!

karlhorky:
Ah, I guess this feature is already (partially) implemented in Chrome:

https://twitter.com/beverloo/status/1167824727009832960

However, they are disabled on macOS due to deficiencies in the macOS notification system (extra buttons for "Site Settings" are not allowed):

https://twitter.com/beverloo/status/1167847717361639425

This is a valid user concern, so I have proposed a few ways of dealing with that, including an extra permissions prompt (1) and requiring an installed PWA (2 and 3).

1. https://twitter.com/karlhorky/status/1167843843238219777
2. https://twitter.com/karlhorky/status/1167848030718046208
3. https://bugs.chromium.org/p/chromium/issues/detail?id=999838

cc @beverloo

xfalcox:
So this been live in Chrome for Windows/ChromeOS/Android for over 6 years, but the update for the spec has been forgot on this PR?

Can we merge this so it will trigger the necessary updates in affects specs like https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#protocol that still doesn't have support for the `type` attribute in the NotificationAction class ?

PR #148, opened 2019-07-16
Review Draft Publication: July 2019

annevk:
A Review Draft for this Workstream will be published shortly, by merging this pull request.

Under the [WHATWG IPR Policy](https://whatwg.org/ipr-policy), Participants may, within 45 days after publication of a Review Draft, exclude certain Essential Patent Claims from the Review Draft Licensing Obligations. See the [IPR Policy](https://whatwg.org/ipr-policy) for details.

### DIFF
--- a/review-drafts/2019-07.bs
+++ b/review-drafts/2019-07.bs
@@ -1,0 +1,1176 @@
+<pre class=metadata>
+Group: WHATWG
+Date: 2019-07-16
+H1: Notifications API
+Shortname: notifications
+Text Macro: TWITTER notifyapi
+Abstract: This standard defines an API to display notifications to the end user, typically outside the top-level browsing context's viewport. It is designed to be compatible with existing notification systems, while remaining platform-independent.
+Translation: ja https://triple-underscore.github.io/notifications-ja.html
+Translation: zh-cn https://w3c-html-ig-zh.github.io/notifications/whatwg/
+</pre>
+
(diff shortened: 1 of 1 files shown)


annevk:
I guess we don't have to publish this as there haven't been edits since the last draft?

foolip:
The previous one was https://notifications.spec.whatwg.org/review-drafts/2019-01/ and that's the date of https://notifications.spec.whatwg.org/, so yeah, publishing seems like it wouldn't do any good.

annevk:
So the one argument in favor of publishing despite nothing having changed since the last Review Draft is that https://whatwg.org/workstream-policy#review-drafts suggests it happens every six months.

@whatwg/sg should probably make the call.

henceproved:
I don't think we should publish another review draft, if nothing has changed. That doesn't seem like a good use of the editors' time. The policy states "approximately six months", but the SG will discuss to see if we should just add a note to the policy saying the review draft publishing will be skipped in such cases.

michaelchampion:
Agree with @henceproved … So long as it’s recorded somewhere (like this GH repo) that there was no Review draft because there were no changes to review for possible patent exclusions, no need to busy work just because of the letter (not intent!) of the workstream policy.  Yes, we should update the policy to make this more clear.

annevk:
Filed https://github.com/whatwg/sg/issues/96 for that clarification.

PR #167, opened 2021-04-22
Editorial: align with modern IDL prose conventions

annevk:
***
<a href="https://whatpr.org/notifications/167.html" title="Last updated on Apr 26, 2021, 6:25 AM UTC (cd5dd5c)">Preview</a> | <a href="https://whatpr.org/notifications/167/d713f3b...cd5dd5c.html" title="Last updated on Apr 26, 2021, 6:25 AM UTC (cd5dd5c)">Diff</a>

PR #168, opened 2021-04-22
Get rid of entry settings object usage

### LINKED ISSUE #86: Remove usage of "entry settings object"
It seems most can be replaced with "relevant settings object" easily, but not entirely sure about usage in https://notifications.spec.whatwg.org/#create-a-notification.

annevk:
Use the relevant settings object instead.

Fixes #86.

(Note that this builds upon #167.)

***
<a href="https://whatpr.org/notifications/168.html" title="Last updated on Apr 26, 2021, 6:46 AM UTC (c7f1660)">Preview</a> | <a href="https://whatpr.org/notifications/168/a3e44ea...c7f1660.html" title="Last updated on Apr 26, 2021, 6:46 AM UTC (c7f1660)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -9,11 +9,19 @@ Translation: ja https://triple-underscore.github.io/notifications-ja.html
 Translation: zh-cn https://w3c-html-ig-zh.github.io/notifications/whatwg/
 </pre>
 
-<!-- Remove this once https://github.com/whatwg/html/pull/6360 is resolved. -->
+<!-- Remove the HTML entry once https://github.com/whatwg/html/pull/6360 is resolved. -->
 <pre class=anchors>
 urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html#; spec: html;
     type: dfn;
         text: DOM manipulation task source; url: dom-manipulation-task-source
+urlPrefix: https://w3c.github.io/vibration/
+    urlPrefix: #dfn-; type: dfn
+        text: perform vibration
+        text: validate and normalize
+    urlPrefix: #idl-def-; type: interface
+        text: VibratePattern; url: vibratepattern
+urlPrefix: https://tc39.github.io/ecma262/#sec-object.; type: dfn
+    text: freeze
 </pre>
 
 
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -150,114 +158,109 @@ clipped corners.
...
+ <var>options</var>["{{NotificationOptions/body}}"].
 
-  <li><p>Set <var>notification</var>'s <a for=notification>tag</a> to
-  <var>options</var>["{{NotificationOptions/tag}}"].
+ <li><p>Set <var>notification</var>'s <a for=notification>tag</a> to
+ <var>options</var>["{{NotificationOptions/tag}}"].
 
-  <li><p>Let <var>baseURL</var> be the API base URL specified by the
-  <a>entry settings object</a>. <span class=XXX>Or incumbent?</span>
+ <li><p>Let <var>baseURL</var> be the API base URL specified by <var>settings</var>.
```
domenic, line 206: This could be `<var>settings</var>'s <a for="environment settings object">API base URL</a>`

notifications.bs
```diff
@@ -718,7 +722,7 @@ constructor steps are:
 <h3 id=static-members>Static members</h3>
 
 <p>The static <dfn attribute for=Notification><code>permission</code></dfn> getter steps are to
-return the <a>permission</a> for the <a>entry settings object</a>'s
+return the <a>permission</a> for <a>this</a>'s <a>relevant settings object</a>'s
```
domenic, line 725: This is static so it doesn't work. You want the `<a>current settings object</a>`.

notifications.bs
```diff
@@ -696,7 +699,8 @@ constructor steps are:
  <a>throw</a> a {{TypeError}}.
 
  <li><p>Let <var>notification</var> be the result of <a>creating a notification</a> given
- <var>title</var> and <var>options</var>. Rethrow any exceptions.
+ <var>title</var>, <var>options</var>, and <a>this</a>'s <a>relevant settings object</a>.
```
domenic, line 702: They're equivalent in constructors, but using this's relevant settings object here and current settings object a few lines up is a bit confusing.

notifications.bs
domenic: Hmm this probably is preexisting in requestPermission() somehow.

domenic (changes requested):
Statics are problematic

notifications.bs
annevk: I fixed both, thanks for spotting that!

annevk:
New commit message:
```
Get rid of entry settings object usage

Use the relevant global/settings object where possible; use the current global/settings object in statics.

Fixes #86.
```

domenic (approved):

PR #175, opened 2021-11-04
Fixes #93 - Restrict requesting permission to secure contexts

### LINKED ISSUE #93: Restrict the Notifications API to secure contexts
Chromium is moving to [restrict access to the notifications API over insecure contexts](https://groups.google.com/a/chromium.org/forum/#!topic/Blink-dev/IVgkxkRNtMo). There will be a deprecation warning in M58, scheduled for April 2017, with full removal in M60, scheduled for August 2017. The proposed behaviour is that calls to Notification.requestPermission() and the Notification constructor from an insecure context will act as though the user has denied permission (similar to the behaviour of the geolocation API from an insecure context).

Motivations for the restriction are as follows:

 * The notification permission acceptance rate for HTTPS sites is 3x greater than for HTTP sites in Chrome
 * Over 90% of daily notification permission requests in Chrome are already made from a secure context
 * Notifications are overwhelming produced in tandem with the Push API which is [already restricted to secure contexts](https://w3c.github.io/push-api/#security-and-privacy-considerations). This restriction is only relevant for sites displaying local notifications while they are open in a browser tab
 * Aligning with the requirements of the Push API reduces developer friction and improves platform predictability
 * Chromium already requires local notifications on Android to be requested over a secure context
 * Notifications allow access to privileged system UI, making them a powerful feature

In light of this, we'd like to propose an update to the spec to explicitly require that user agents only make the Notifications API available over secure contexts, and to codify the behaviour of the API when requested from an insecure context. We also welcome suggestions on our proposed behaviour and whether it could be improved.

miketaylr:
- [x] At least two implementers are interested (and none opposed):
     * Chrome, Firefox, and Safari behave this way today.
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/31495 landed.
- [x] [Implementation bugs] N/A

***
<a href="https://whatpr.org/notifications/175.html" title="Last updated on Nov 19, 2021, 1:23 AM UTC (b49c447)">Preview</a> | <a href="https://whatpr.org/notifications/175/df1b6db...b49c447.html" title="Last updated on Nov 19, 2021, 1:23 AM UTC (b49c447)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -317,6 +317,14 @@ meaning "<code>granted</code>". In that case
 "<code>granted</code>" is simply returned as there would be no reason
 for the application to ask for <a>permission</a>.
 
+<h3 id="permissions-integration">Permissions integration</h3>
+
+<p>The Notifications API is a <a>powerful feature</a> which is identified by the string
+"notifications" and has its <a for="powerful feature">allowed in non-secure contexts</a> flag set.
+[[!Permissions]]
+
+<p class="note">This specification's "<code>default</code>" <a>permission</a>
+state maps to the [=permission/prompt=] permission state in the Permissions API.
 
 <h3 id=direction>Direction</h3>
 
@@ -729,8 +737,20 @@ constructor steps are:
 return the <a>permission</a> for the <a>current settings object</a>'s
 <a for="environment settings object">origin</a>.
 
-<p class=note>If you edit standards please refrain from copying the above. Synchronous
-permissions are like synchronous IO, a bad idea.
+<div class=note>
+ <p>If you edit standards please refrain from copying the above. Synchronous permissions are like
+ synchronous IO, a bad idea.
+
+ <p>Developers are encouraged to use the Permissions {{Permissions/query()}} method
+ instead. [[Permissions]]
+
+ <pre class="example" id="permissions-query-example">
+    const permission = await navigator.permissions.query({name: "notifications"});
+    if (permission.state === "granted") {
+       // We have permission to use the API…
+    }
+ </pre>
+</div>
 
 <p>The static
 <dfn method for=Notification><code>requestPermission(<var>deprecatedCallback</var>)</code></dfn>
@@ -739,10 +759,12 @@ method steps are:
 <ol>
  <li><p>Let <var>global</var> be the <a>current global object</a>.
 
- <li><p>Let <var>origin</var> be the <a>current settings object</a>'s
- <a for="environment settings object">origin</a>.
- <!-- Arguably this should snapshot origin somehow so that changes to it do not impact the
-      in parallel part. -->
+ <li><p>Let <var>permissionDescriptor</var> be the {{PermissionDescriptor}} with
+ {{PermissionDescriptor/name}} set to "<code>notifications</code>".
+
+ <li><p>Let <var>permissionStatus</var> be the result of
+ <a lt="create a PermissionStatus">creating a `PermissionStatus`</a> for
+ <var>permissionDescriptor</var>.
 
  <li><p>Let <var>promise</var> be <a for=/>a new promise</a> in <a>this</a>'s <a>relevant Realm</a>.
 
@@ -750,24 +772,25 @@ method steps are:
   <p>Run these steps <a>in parallel</a>:
 
   <ol>
-   <li><p>Let <var>permission</var> be <a>permission</a> for <var>origin</var>.
+   <li><p>Run the <a for="powerful feature">permission query algorithm</a> with
+   <var>permissionDescriptor</var> and <var>permissionStatus</var>.
 
-   <li><p>If <var>permission</var> is "<code>default</code>", then ask the user whether showing
-   notifications for <var>origin</var> is acceptable. If it is, then set <var>permission</var> to
-   "<code>granted</code>"; otherwise "<code>denied</code>".
+   <li><p>Let <var>permissionState</var> be <var>permissionStatus</var>'s
+   {{PermissionStatus/state}}.
+
+   <li><p>If <var>permissionState</var> is {{PermissionState/"prompt"}}, then set
+   <var>permissionState</var> to "<code>default</code>".
 
    <li>
     <p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given <var>global</var>
     to run these steps:
 
     <ol>
-     <li><p>Set <a>permission</a> for <var>origin</var> to <var>permission</var>.
-
      <li><p>If <var>deprecatedCallback</var> is given, then <a for=/>invoke</a>
-     <var>deprecatedCallback</var> with « <var>permission</var> ». If this throws an exception, then
-     <a>report the exception</a>.
+     <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an
+     exception, then <a>report the exception</a>.
 
-     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permission</var>.
+     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
     </ol>
   </ol>
 
@@ -1186,6 +1209,7 @@ Michael Cooper,
 Michael Henretty,
 Michael™ Smith,
 Michael van Ouwerkerk,
+Mike Taylor,
 Nicolás Satragno,
 Olli Pettay,
 Peter Beverloo,
--- a/notifications.bs
+++ b/notifications.bs
@@ -325,6 +325,7 @@ for the application to ask for <a>permission</a>.
 <p class="note">This specification's "<code>default</code>" <a>permission</a>
 state maps to the [=permission/prompt=] permission state in the Permissions API.
 
+
 <h3 id=direction>Direction</h3>
 
 <p>This section is written in terms equivalent to those used in the Rendering
--- a/notifications.bs
+++ b/notifications.bs
@@ -320,8 +320,7 @@ for the application to ask for <a>permission</a>.
 <h3 id="permissions-integration">Permissions integration</h3>
 
 <p>The Notifications API is a <a>powerful feature</a> which is identified by the string
-"notifications" and has its <a for="powerful feature">allowed in non-secure contexts</a> flag set.
-[[!Permissions]]
+"notifications". [[!Permissions]]
 
 <p class="note">This specification's "<code>default</code>" <a>permission</a>
 state maps to the [=permission/prompt=] permission state in the Permissions API.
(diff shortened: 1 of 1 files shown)


miketaylr:
@annevk mind looking at https://github.com/whatwg/notifications/pull/175/commits/afd99ad797c151e1d0bc6644b3ab808c48689121?

notifications.bs
```diff
@@ -782,20 +786,32 @@ method steps are:
...
+   <li>Return the result of <a>resolving a permission request promise</a> with
+   <var>permissionState</var> and <var>deprecatedCallback</var>, if given.
+  </ol>
+</ol>
 
-    <ol>
-     <li><p>If <var>deprecatedCallback</var> is given, then <a for=/>invoke</a>
-     <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an
-     exception, then <a>report the exception</a>.
+To <dfn lt="resolving a permission request promise">resolve a permission request promise</dfn>
```
annevk, line 794: Nit: use explicit `<p>`.

notifications.bs
```diff
@@ -767,13 +769,15 @@ method steps are:
...
 
- <li><p>Let <var>promise</var> be <a for=/>a new promise</a> in <a>this</a>'s <a>relevant Realm</a>.
-
  <li>
   <p>Run these steps <a>in parallel</a>:
 
   <ol>
-   <li><p>Run the <a for="powerful feature">permission query algorithm</a> with
+   <li><p>If <var>global</var>'s associated <a>environment settings object</a> is not a
+   <a>secure context</a>, return the result of <a>resolving a permission request promise</a> with
```
annevk, line 777: I think you need to keep the existing setup here because you are returning a promise from in parallel steps, but that does not end up returning anything to the initial caller. It returns it into the in parallel void. 😊

annevk (commented):
Mainly nits and one structural issue. Thanks for fixing this! 😊

miketaylr:
Thanks for the super useful feedback, think I have it all addressed. 

r? @annevk

notifications.bs
```diff
@@ -782,20 +789,31 @@ method steps are:
...
-     <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an
-     exception, then <a>report the exception</a>.
+  <li><p>Return <var>promise</var>.
+</ol>
 
-     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
-    </ol>
-  </ol>
+<p>To <dfn>resolve a permission request</dfn> given promise <var>promise</var>,
+[=permission state=] <var>permissionState</var>, and optional callback function
```
miketaylr, line 800: My first PR had the type for permissionState as [`{{PermissionStatus/state}}`](https://w3c.github.io/permissions/#state-attribute), but I think that was a bug. Here it's a [`[=permission state=]`](https://w3c.github.io/permissions/#dfn-permission-state), that way we can pass in "denied" explicitly for the non-secure branch before we even have a PermissionStatus instance. (And `{{PermissionStatus/state}}` is ultimately set to `[=permission state=]` after the permission query algorithm runs).

annevk: I'm not sure this works since permission state belongs to something else. Perhaps this should just be a string?

notifications.bs
```diff
@@ -734,9 +734,14 @@ constructor steps are:
...
-return the <a>permission</a> for the <a>current settings object</a>'s
-<a for="environment settings object">origin</a>.
+<p>The static <dfn attribute for=Notification><code>permission</code></dfn> getter steps are:
+
+<ol>
+ <li><p>Let <var>settings</var> be the <a>current settings object</a>.
+
+ <li><p>If <var>settings</var> is a <a>secure context</a>, then return the <a>permission</a> for
+ <var>settings</var>'s <a for="environment settings object">origin</a>. Otherwise, return
+ "<code>denied</code>".
```
annevk, line 744: This is weird as actually we should run the permission query algorithm here too, no? And change prompt to default? And remove section that defines `<a>permission</a>` inline.

notifications.bs
miketaylr: Yeah, string makes more sense. Will do.

notifications.bs
miketaylr: OK, let's see if we can make a common <dfn> for both to use.

miketaylr:
OK, let's try again!

notifications.bs
```diff
@@ -735,8 +733,32 @@ constructor steps are:
...
-return the <a>permission</a> for the <a>current settings object</a>'s
-<a for="environment settings object">origin</a>.
+return the result of <a>querying the notifications permission state</a>.
+
+<p>To <dfn>query the notifications permission state</dfn>, run these steps:
+
+<ol>
+ <li><p>Let <var>settings</var> be the <a>current settings object</a>.
+
+ <li><p>If <var>settings</var> is not a <a>secure context</a>, return "<code>denied</code>".
```
annevk, line 743: Nit: then return*

notifications.bs
```diff
@@ -735,8 +733,32 @@ constructor steps are:
...
+
+ <li><p>Otherwise, let <var>permissionDescriptor</var> be the {{PermissionDescriptor}} with
```
annevk, line 745: Nit: no need for otherwise since it returns.

notifications.bs
```diff
@@ -735,8 +733,32 @@ constructor steps are:
...
+ {{PermissionDescriptor/name}} set to "<code>notifications</code>".
+
+ <li><p>Let <var>permissionStatus</var> be the result of
+ <a lt="create a PermissionStatus">creating a `PermissionStatus`</a> for
+ <var>permissionDescriptor</var>.
+
+ <li><p>Run the <a for="powerful feature">permission query algorithm</a> with
+ <var>permissionDescriptor</var> and <var>permissionStatus</var>.
+
+ <li><p>Let <var>permissionState</var> be <var>permissionStatus</var>'s {{PermissionStatus/state}}.
```
annevk, line 755: Ideally we'd return an internal concept here and not a property, but I guess that doesn't really exist?

notifications.bs
```diff
@@ -735,8 +733,32 @@ constructor steps are:
...
+
+ <li><p>If <var>permissionState</var> is {{PermissionState/"prompt"}}, then set
+ <var>permissionState</var> to "<code>default</code>".
```
annevk, line 758: Maybe this should just return "default"?

notifications.bs
```diff
@@ -760,42 +782,28 @@ method steps are:
...
-     exception, then <a>report the exception</a>.
-
-     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
-    </ol>
+    <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
+   </ol>
   </ol>
 
- <li><p>Return <var>promise</var>.
+  <li><p>Return <var>promise</var>.
```
annevk, line 806:
```suggestion
 <li><p>Return <var>promise</var>.
```

notifications.bs
```diff
@@ -760,42 +782,28 @@ method steps are:
...
-
-   <li><p>Let <var>permissionState</var> be <var>permissionStatus</var>'s
-   {{PermissionStatus/state}}.
+   <li><p>Let <var>permissionState</var> be the result of
+   <a>querying the notifications permission state</a>.
 
-   <li><p>If <var>permissionState</var> is {{PermissionState/"prompt"}}, then set
-   <var>permissionState</var> to "<code>default</code>".
+   <li><p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given
+   <var>global</var> to run these steps:
```
annevk, line 795:
```suggestion
   <li>
    <p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given <var>global</var> 
    to run these steps:
```

notifications.bs
```diff
@@ -760,42 +782,28 @@ method steps are:
...
 
-   <li>
-    <p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given <var>global</var>
-    to run these steps:
+   <ol>
+    <li><p>If <var>deprecatedCallback</var> is given, then <a for=/>invoke</a>
+    <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an
+    exception, then <a>report the exception</a>.
```
annevk, line 800:
```suggestion
    <ol>
     <li><p>If <var>deprecatedCallback</var> is given, then <a for=/>invoke</a> 
     <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an exception, 
     then <a>report the exception</a>.
```

notifications.bs
```diff
@@ -760,42 +782,28 @@ method steps are:
...
 
-    <ol>
-     <li><p>If <var>deprecatedCallback</var> is given, then <a for=/>invoke</a>
-     <var>deprecatedCallback</var> with « <var>permissionState</var> ». If this throws an
-     exception, then <a>report the exception</a>.
-
-     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
-    </ol>
+    <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
+   </ol>
```
annevk, line 803:
```suggestion
     <li><p><a for=/>Resolve</a> <var>promise</var> with <var>permissionState</var>.
    </ol>
```

annevk (commented):
This is a really nice refactoring! Mainly formatting nits left.

notifications.bs
miketaylr: (please let me know when annevk-lint is available for vscode!)

notifications.bs
miketaylr: Yeah, good point.

notifications.bs
miketaylr: We could do something like `if {{PermissionStatus/state}} is "granted", then return "granted"`, for granted and denied . Or just leave it?

miketaylr:
Bah, somehow I messed up a rebase. Let me do some surgery.

miketaylr:
> Bah, somehow I messed up a rebase. Let me do some surgery.

OK, everything looks fine. But I've learned that I am not good at merging in GitHub suggestions after I've made more local branch changes. 🙈 

PTAL @annevk

notifications.bs
annevk: This is a problem with the Permissions specification in that it operates on IDL attributes rather than internal state. Invoking IDL attributes directly is problematic as they could be overridden by script in theory so it's not what you can actually implement.

I don't think we can fix this at this point until Permissions is better about this. @marcoscaceres is this an issue that's being tracked?

notifications.bs
annevk: This got lost in the rebase mixup I'm afraid.

notifications.bs
```diff
@@ -320,13 +320,11 @@ for the application to ask for <a>permission</a>.
 <h3 id="permissions-integration">Permissions integration</h3>
 
 <p>The Notifications API is a <a>powerful feature</a> which is identified by the string
-"notifications" and has its <a for="powerful feature">allowed in non-secure contexts</a> flag set.
-[[!Permissions]]
+"notifications". [[!Permissions]]
 
 <p class="note">This specification's "<code>default</code>" <a>permission</a>
 state maps to the [=permission/prompt=] permission state in the Permissions API.
```
annevk, line 327:
```suggestion


```

annevk (commented):
It seems some feedback got undone due to the rebase mixup. I think it would be nice if we also removed https://whatpr.org/notifications/175.html#permission-model as part of this and changed https://whatpr.org/notifications/175.html#constructors and https://whatpr.org/notifications/175.html#dom-serviceworkerregistration-shownotification to also use this new query operation you created. And perhaps the query operation itself should move to https://whatpr.org/notifications/175.html#permissions-integration?

miketaylr:
> It seems some feedback got undone due to the rebase mixup.

Yep, I'm declaring bankruptcy here and will open a new PR.

PR #176, opened 2021-11-19
Restrict Notifications to secure contexts.

### LINKED ISSUE #93: Restrict the Notifications API to secure contexts

miketaylr:
Continuing the mess I made in #175.

Fixes #93

***
<a href="https://whatpr.org/notifications/176.html" title="Last updated on Nov 22, 2021, 7:57 PM UTC (4eff822)">Preview</a> | <a href="https://whatpr.org/notifications/176/df1b6db...4eff822.html" title="Last updated on Nov 22, 2021, 7:57 PM UTC (4eff822)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -294,13 +294,13 @@ removed from the <a>list of notifications</a>.
 <h3 id="permissions-integration">Permissions integration</h3>
 
 <p>The Notifications API is a <a>powerful feature</a> which is identified by the
-<a for="powerful feature">name</a> "`notifications`". [[!Permissions]]
+<a for="powerful feature">name</a> "<code>notifications</code>". [[!Permissions]]
 
 <p>To <dfn>get the notifications permission state</dfn>, run these steps:
 
 <ol>
  <li><p>Let <var>permissionState</var> be the result of <a>getting the current permission state</a>
- with "`notifications`".
+ with "<code>notifications</code>".
 
  <li><p>If <var>permissionState</var> is "<code>prompt</code>", then return "<code>default</code>".
 
@@ -747,7 +747,7 @@ method steps are:
 
   <ol>
    <li><p>Let <var>permissionState</var> be the result of
-   <a>getting the notifications permission state</a>.
+   <a>requesting permission to use</a> {{PermissionName/notifications}}.
 
    <li>
     <p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given <var>global</var>
--- a/notifications.bs
+++ b/notifications.bs
@@ -706,9 +706,9 @@ constructor steps are:
   <p>Run these steps <a>in parallel</a>:
 
   <ol>
-   <li><p>If <a>permission</a> for <var>notification</var>'s <a for=notification>origin</a> is not
-   "<code>granted</code>", then <a>queue a task</a> to <a>fire an event</a> named <code>error</code>
-   on <a>this</a>, and abort these steps.
+   <li><p>If the result of <a>getting the notifications permission state</a> is not
+   "<code>granted</code>", then <a>queue a task</a> to <a>fire an event</a> named
+   <code>error</code> on <a>this</a>, and abort these steps.
 
    <li><p>Run the <a>fetch steps</a> for <var>notification</var>.
 
--- a/notifications.bs
+++ b/notifications.bs
@@ -297,6 +297,17 @@ removed from the <a>list of notifications</a>.
 "notifications" and has its <a for="powerful feature">allowed in non-secure contexts</a> flag set.
 [[!Permissions]]
 
+<p>To <dfn>get the notifications permission state</dfn>, run these steps:
+
+<ol>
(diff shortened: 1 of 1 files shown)


miketaylr:
(this will be even simpler with https://github.com/w3c/permissions/commit/a2464fbc53b4a114e15a923b05bd589444038856)

miketaylr:
PTAL @annevk @marcoscaceres

notifications.bs
```diff
@@ -291,37 +291,21 @@ removed from the <a>list of notifications</a>.
...
+ <li><p>If <var>permissionState</var> is "<code>prompt</code>", then return "<code>default</code>".
 
-<p>The Notifications API is a <a>powerful feature</a> which is identified by the string
-"notifications" and has its <a for="powerful feature">allowed in non-secure contexts</a> flag set.
-[[!Permissions]]
+ <li><p>Return <var>permissionState</var>.
+</ol>
 
 <p class="note">This specification's "<code>default</code>" <a>permission</a>
 state maps to the [=permission/prompt=] permission state in the Permissions API.
```
annevk, line 311: Let's remove this as well, especially no we longer define permission.
```suggestion
```

annevk (approved):
This looks great. Just a couple of nits.

annevk:
I tried to think through this again. How come `Notification.permission` does not end up prompting, but `Notification.requestPermission()` does? Wouldn't that require passing some kind of flag when obtaining the permission? The other places shouldn't end up prompting either, by the way.

notifications.bs
```diff
@@ -291,40 +291,21 @@ removed from the <a>list of notifications</a>.
...
-  <dt>"<code>denied</code>"
-  <dd><p>This means the user does not want
-  <a>notifications</a>.
+<h3 id="permissions-integration">Permissions integration</h3>
 
-  <dt>"<code>granted</code>"
-  <dd><p>This means <a>notifications</a> can be
-  displayed.
-</dl>
+<p>The Notifications API is a <a>powerful feature</a> which is identified by the
```
annevk, line 296:
```suggestion
<p>The Notifications API is a <a>powerful feature</a> which is identified by the
```
My bad...

miketaylr: Heh, all good. Just pushed a fix.

miketaylr:
> I tried to think through this again. How come `Notification.permission` does not end up prompting, but `Notification.requestPermission()` does?

Hmm, you're right - this isn't quite what we want. I think we should call https://w3c.github.io/permissions/#dfn-request-permission-to-use directly inside `requestPermission` instead of `get the notifications permission state`. The other places should be fine then.

notifications.bs
```diff
@@ -760,27 +740,14 @@ method steps are:
...
-   <li><p>Run the <a for="powerful feature">permission query algorithm</a> with
-   <var>permissionDescriptor</var> and <var>permissionStatus</var>.
-
-   <li><p>Let <var>permissionState</var> be <var>permissionStatus</var>'s
-   {{PermissionStatus/state}}.
-
-   <li><p>If <var>permissionState</var> is {{PermissionState/"prompt"}}, then set
-   <var>permissionState</var> to "<code>default</code>".
+   <li><p>Let <var>permissionState</var> be the result of
+   <a>requesting permission to use</a> {{PermissionName/notifications}}.
```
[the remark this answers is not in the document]
miketaylr: Yep, it only returns the strings "granted" or "denied":
> This algorithm returns either "granted" or "denied".

annevk (approved):

annevk:
Thanks a lot @miketaylr for getting this cleaned up!

miketaylr:
> Thanks a lot @miketaylr for getting this cleaned up!

Thanks for patiently guiding me through it. :)

PR #206, opened 2024-04-04
Editorial: Add a missing comma

saschanaz:
***
<a href="https://whatpr.org/notifications/206.html" title="Last updated on Apr 4, 2024, 2:47 PM UTC (428832a)">Preview</a> | <a href="https://whatpr.org/notifications/206/fab345b...428832a.html" title="Last updated on Apr 4, 2024, 2:47 PM UTC (428832a)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -451,7 +451,7 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
 
 <ol>
  <li><p>Wait for any <a for=/ lt=fetch>fetches</a> to complete and <var>notification</var>'s
- <a for=notification>image resource</a> <a for=notification>icon resource</a>, and
+ <a for=notification>image resource</a>, <a for=notification>icon resource</a>, and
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -451,7 +451,7 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
 
 <ol>
  <li><p>Wait for any <a for=/ lt=fetch>fetches</a> to complete and <var>notification</var>'s
```
saschanaz, line 453: Out of scope of this PR: This effectively is await-ing the [fetch steps](https://notifications.spec.whatwg.org/#fetch-steps) in bad way; the fetch steps say the resources are set only when supported but here we are waiting for those resources to be set.

Is there a better wait to wait for a parallel task? 🤔

annevk: I think we should probably set "image resource" either way (to a decoded resource or to failure or some such). And then we wait for it to not be in its initial state (null?).

annevk (approved):
Thanks!

Issue #204, opened 2024-02-09
Should a non-persistent notification be auto-closed or not?


saschanaz:
### What is the issue with the Notifications API Standard?

Currently the spec says: https://notifications.spec.whatwg.org/#lifetime-and-ui-integrations

> User agents should run the [close steps](https://notifications.spec.whatwg.org/#close-steps) for a [non-persistent notification](https://notifications.spec.whatwg.org/#non-persistent-notification) a couple of seconds after they have been created.

The actual browser behavior is different:

1. Chrome and Firefox on Windows do not auto-close non-persistent notifications; and instead clicking the notification after tab close just opens the origin URL.
2. Safari and Firefox on macOS/Linux auto-close non-persistent notifications if the navigation happens away from the triggering page

[We had internal argument against the second behavior](https://bugzilla.mozilla.org/show_bug.cgi?id=1879366); non-persistent notification may also include some crucial content that user would like to keep in the notification center. My immediate impression is that such notification should use persistent way instead, but given the browser differences, I think should discuss about the behavior here for better interop.

saschanaz:
Also nobody clears non-persistent notifications from workers 👀

reillyeon:
Chrome and Safari (haven't tested Firefox) violate the next line as well:

> User agents should not display [non-persistent notification](https://notifications.spec.whatwg.org/#non-persistent-notification) in a platform’s "notification center" (if available).

Chrome on Windows and ChromeOS (haven't tested other platforms) add non-persistent notifications to the platform's "notification center". Safari does as well, though as mentioned above it removes them when the tab is navigated or closed. Chrome does not.

annevk:
> just opens the origin URL

From all the things this is the weirdest behavior listed to me, but we could allow this as one of the alternatives. It could use the same vague-ish definition we plan to use for `navigate`. I'm not convinced interop matters much here as this is all UI-related and the only issue about it in over a dozen years has been from a browser engineer.

I'd also be happy to remove or weaken most of the other requirements in this section. I'd probably keep the non-persistent/persistent distinction as-is as it's kinda a pain to rename and it still seems accurate lifetime-wise, at least from the perspective of the web developer.

PR #213, opened 2024-07-10
Allow notifications and actions to specify a navigable URL

annevk:
This makes notifications more declarative by not requiring explicit handling of the clicks by the web application.

---

This change is part of a larger [Declarative Web Push effort](https://github.com/w3c/push-api/issues/360), but can hopefully land in isolation as a small incremental step toward that goal.

- [x] At least two implementers are interested (and none opposed):
   * WebKit
   * Gecko
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * IDL will be tested, but otherwise there's no good infrastructure for this as-of-yet.
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://issues.chromium.org/issues/382298314
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1935325
   * WebKit: N/A
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/mdn/issues/720
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/notifications/213.html" title="Last updated on Aug 1, 2025, 5:45 AM UTC (80e6695)">Preview</a> | <a href="https://whatpr.org/notifications/213/d7f3e1a...80e6695.html" title="Last updated on Aug 1, 2025, 5:45 AM UTC (80e6695)">Diff</a>


notifications.bs
```diff
@@ -211,6 +217,24 @@ string <var>title</var>, {{NotificationOptions}} <a for=/>dictionary</a> <var>op
...
+
+  <ol>
+   <li><p>Let <var>url</var> be the result of <a lt="URL parser">parsing</a>
+   <var>options</var>["{{NotificationOptions/url}}"] using <var>baseURL</var>.
+
+   <li>
+    <p>If <var>url</var> is not failure:
+
+    <ol>
+     <li><p>If <var>url</var>'s <a for=url>origin</a> is not <a>same origin</a> with
```
annevk, line 231: Let's remove this check and the one below as JavaScript can already do whatever.

annevk:
@saschanaz @beverloo thoughts?

saschanaz:
I like the idea in genaral. Can we split editorial refactoring into separate PR though? It's a bit pain in general when I just want to diff the normative changes either for the review or for the impl after merge.

annevk:
There should not be any refactoring? I did the refactoring in 0fd4b82dc104e078b211e925d927d612c90ba0fb.

saschanaz:
Relocating the body definition, reordering name and title, etc etc.

saschanaz:
1 or 2 minor changes are fine but getting more starts to be annoying, especially if we get multiple review rounds... 😛

Should we also fold in some part of #204 here? Non-persistent notification with a URL certainly has a valid action to do even if the page is closed.

saschanaz:
(Interesting that we are currently not telling what action is invoked for non-persistent notifications, per the existing steps)

saschanaz:
A parallel question, is WebKit going to implement actions?

notifications.bs
```diff
@@ -211,6 +217,11 @@ string <var>title</var>, {{NotificationOptions}} <a for=/>dictionary</a> <var>op
  <li><p>Set <var>notification</var>'s <a for=notification>body</a> to
  <var>options</var>["{{NotificationOptions/body}}"].
 
+ <li><p>If <var>options</var>["{{NotificationOptions/url}}"] <a for=map>exists</a>, then
+ <a lt="URL parser">parse</a> it using <var>baseURL</var>, and if that does not return failure, set
```
saschanaz, line 221: Should we ban certain kinds of URLs, e.g. non-HTTP/S ones? What about cross origin URLs? (Technically the latter is allowed when using event listeners, though.)

annevk:
- The couple nits have now landed separately (along with some other nits) and I rebased this.
- I'd prefer fixing #204 separately. I don't really have an opinion on how that should work.
- WebKit is interested in notification actions.

saschanaz:
Some extra questions: 

1. Should we still give a chance for the webpage to handle this, so that existing active tab can e.g. open notification inbox without opening a new tab? Doing that might be a bit tricky though:

    1. Tab A triggers a notification with URL
    2. Tab A goes away
    3. User opens tab B and C
    4. User clicks notification
    5. Now which tab should respond? 🤷‍♀️

    cc @smaug---- 

2. We silently ignore URL parsing error here (probably for consistency with other fields), but should we throw explicitly for this one? Clicking notification with incorrect URL wouldn't open anything, and such thing probably should fail early without being shown.

annevk:
1. We could potentially offer that in a future version, but I don't think we need to for v1. Let's see how much adoption there will be for Declarative Web Push first.
2. Hmm, per https://github.com/whatwg/notifications/commit/0f8a7b651c5e7afaae6172ffc0b759e70af32a9a it seems historically we didn't really define what happened on failure and just copy-and-pasted our way forward from there (with "minor" changes along the way such as adopting the WHATWG URL parser). I suppose we could start breaking with that precedent now and start throwing for new URL members. Not sure I lean particularly strongly either way.

smaug----:
Having a way to bound to a tab might affect the API usage quite a bit.  Needing to open a tab all the time even if the site was already open would be a bit annoying.

But perhaps you've had some other types of use cases in mind. In which way do you imagine one to use only the URL + new top level navigable?

annevk:
This feature is part of https://github.com/w3c/push-api/issues/360 to enable fully declarative push notifications. Perhaps it should not be allowed for `new Notification()`, but it also seemed kinda harmless. And perhaps we should also offer a variant that ends up navigating an existing tab instead, if any, but that seems quite a bit more involved.

cc @beidson

saschanaz:
We should probably also consider not always opening a new tab for declarative push, btw.

annevk:
Yeah we'll think about it and are of course also happy to consider concrete proposals. I suspect that anything we decide to do can easily go on top of this, so I'm not sure I'd consider it blocking.

annevk:
Brady pointed out to me that we already reuse an existing tab if there is one. I don't think HTML has a good primitive for that, but I'll see about adjusting the wording to allow for that. Thanks for pointing it out! (I guess that raises the question as to whether there should be a feature to force a new window, but that's probably best deferred.)

saschanaz:
Hmmmm, In that case I would be interested to learn more about the tab selection algorithm.

beverloo:
Hey Anne, this seems good to us. I've two thoughts -

1. I don't _love_ the `url` property name, but I can't really come up with anything much better either. Using either `navigationUrl` or `activationUrl` is much longer, and inconsistent with other properties on the notification objects.
1. +1 to @saschanaz that some sort of "how to open this" policy is necessary. We see three categories, (1) always open a new tab, (2) re-use an existing tab, and (3) re-use an existing tab unless it's been modified, e.g. dirty form fields.

Perhaps `navigationUrl` is an option to consider as it would work with `navigationPolicy`, which then could then be an enum of sorts. Doesn't have to be for this PR either, and developer feedback might highlight additional options to consider.

beverloo:
Also - we don't have good metrics on use between `Clients.openWindow()` and `WindowClient.navigate()` (or `WindowClient.focus()`) unfortunately, which I can reintroduce but it'll be a while before we have data.

annevk:
I think the policy we end up using in Safari is roughly reuse a tab with the same URL. There's a number of corner cases to imagine where such a vague notion could fall apart, but I'm also not sure we should try to nail it down in detail as I think we want this to be a little flexible to allow changes to it in the future and also because it fundamentally is also part of the overall end user experience.

I could see renaming `url` to `navigate` so it's analogous with the Navigation API. @domenic thoughts?

domenic:
> I could see renaming `url` to `navigate` so it's analogous with the Navigation API. @domenic thoughts?

No strong feelings, but that (or `navigationURL`) seems reasonable.

For the internal slot, it does seem that something more specific than "URL" would be helpful for clarity.

notifications.bs
```diff
@@ -640,6 +682,7 @@ interface Notification : EventTarget {
   readonly attribute NotificationDirection dir;
   readonly attribute DOMString lang;
   readonly attribute DOMString body;
+  readonly attribute USVString navigate;
```
saschanaz, line 685: This kinda sounds like a boolean name, navigateTo maybe?

annevk: It's the same term we use for the Navigation API.

asutherland:
> I think the policy we end up using in Safari is roughly reuse a tab with the same URL. There's a number of corner cases to imagine where such a vague notion could fall apart, but I'm also not sure we should try to nail it down in detail as I think we want this to be a little flexible to allow changes to it in the future and also because it fundamentally is also part of the overall end user experience.

Now that URLPattern is an available primitive, one could imagine letting the site provide a (bounded) list of URLPatterns in precedence order to help with the tab selection.

saschanaz:
Might want to block some tabs from being navigated, in case the state can't be recovered there. (Some form with no backing storage, or some in-progress page either for authentication or payment, etc.)

annevk:
Those seem like reasonable ideas, but I think we should only consider adding complexity once we see significant web developer demand for more fine-grained control.

saschanaz:
```
interface NotificationOptions {
  navigate: {
    url: (URL),
    onExistingTab: (boolean) // initially, and in the future give URLPattern?
  } 
}
```

Might make sense to make `navigate` a nested dictionary in case more options come.

annevk:
That seems a little premature? We could always overload once we establish we need the complexity, no? "Baby steps" and all that.

saschanaz:
If we are going to only ship one extra member, sure. I wonder we should make the tab selection optional if we are going to do an implementation detailed magic there though.

notifications.bs
```diff
@@ -539,14 +555,48 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
+  <p>If <var>navigationURL</var> is non-null:
+
+  <ol>
+   <li><p>If there is an existing <a for=/>top-level traversable</a> <var>traversable</var> within
+   the user agent's <a for="user agent">top-level traversable set</a> whose
+   <a for="navigable">active document</a>'s <a for=Document>URL</a> <a for=url>equals</a>
+   <var>navigationURL</var> with <a for=url/equals><i>exclude fragments</i></a> set to true, then
+   <a for=/>navigate</a> <var>traversable</var> to <var>navigationURL</var> and return. If there are
+   multiple <a for=/>top-level traversables</a> that satisfy the condition the user agent has to
+   pick one in an <a>implementation-defined</a> manner.
```
saschanaz, line 584: Given "losing the data by accidental navigation" is an existing problem with existing solution, Mozilla is against letting this fully implementation defined. We prefer to giving some basic heuristic to prevent it, unless there's a proof that selecting random traversable will not cause a problem.

Here's our suggested revision:

1. Let |traversables| be the user agent’s top-level traversable set whose active document’s URL equals |navigationURL| with exclude fragments set to true. (This part is copied from the PR, but see the above comment)
2. Let |selected| be null.
3. For each |traversable| in |traversables|,
   1. If |traversable| needs beforeunload, continue.
   2. Let |activeDocument| be |traversable|'s active document.
   3. Let |focusState| be the result of running the [has focus steps] with |activeDocument| as the argument.
   4. If |focusState| is true, let |selected| be |traversable| and break.
   5. If |activeDocument|'s visibility state is "visible", let |selected| be |traversable| and break.
4. If |selected| is null and |traversables| is not empty, let |selected| be the first item of |traversables|.
5. If |selected| is null,
   1. Create a fresh top-level traversable given |navigationURL| and return.
6. Navigate |selected| to |navigationURL|.
7. Run the focusing steps with |selected|.

notifications.bs
```diff
@@ -539,14 +555,48 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
+  providing more flexibility to the web developer.
+
+ <li>
+  <p>If <var>navigationURL</var> is non-null:
+
+  <ol>
+   <li><p>If there is an existing <a for=/>top-level traversable</a> <var>traversable</var> within
+   the user agent's <a for="user agent">top-level traversable set</a> whose
+   <a for="navigable">active document</a>'s <a for=Document>URL</a> <a for=url>equals</a>
+   <var>navigationURL</var> with <a for=url/equals><i>exclude fragments</i></a> set to true, then
```
saschanaz, line 581: Wouldn't this always compare the full path? Should we have some base path to compare instead? Otherwise I think each click would effectively always open a new tab unless the user already has a tab for the exact same URL.

Having a base URL would be useful for Slack example, as the URL is `https://app.slack.com/client/(workspace)/(channel)`, and the notification would pass `https://app.slack.com/client/(workspace)` as the base url to only match the same workspace.

notifications.bs
annevk: If the website is not robust against navigation, wouldn't they simply mint a new URL?

notifications.bs
annevk: I think Slack would probably opt to handle the notification in its service worker if there is one, using this as fallback. Perhaps over time we can address more use cases here, but having some deployment experience first would help a lot with that.

saschanaz:
Now I'm confused, I was assuming that the new intention is to reuse the existing tabs, and the full URL comparison was accidental. It seems the intention is actually to reuse only if the URL is exactly the same of one of the existing tabs and open new tab for anything else. Not sure how that's strictly useful than just open new tab for everything for V1, given a lot of existing use cases assign a new URL for each notification because a notification frequently means there's a new content (e.g. Mastodon, Twitter, YouTube, Slack, Instagram, etc.. Chat apps may have higher frequency of having same URL but still with severely limited chance.)

annevk:
The full URL comparison is not accidental. On all Apple platforms, when you click a link outside the default browser (or run `open https://github.com/whatwg/notifications/pull/213`), it'll open the default browser, and if that browser is Safari, it'll navigate an existing tab with that URL.

That seems like reasonable behavior for external links as part of v1 of this feature, but as other behavior is reasonable too and this is really the realm of user interfaces, the PR doesn't require an exact behavior, just that a navigation occurs.

As we get more adoption and feature requests around this, we can certainly add some enhancements to this, such as your proposed `click` event, but none of that seems necessary to us for v1.

notifications.bs
saschanaz: I'm still not convinced why the specced behavior is useful. It looks either not useful as websites with permalinks will always get new tabs, or surprising as websites without permalinks navigate within existing tab unexpectedly.

Given Apple shipped the feature, maybe you can share the idea why Apple decided this way or proof that this doesn't cause footgun.

saschanaz:
>since you commented on a resolved thread without reopening it:

I couldn't reopen it, the unresolve button doesn't work for me, sorry. And I did not resolve it, so I wonder you can?

>The behavior we ship aligns with how links from elsewhere in the OS cause the browser to navigate today.

AH. AAH. I was confused about this statement while sitting in front of a Windows machine and then started thinking maybe this is macOS convention. And boom, that's true, if I click links on Mail app with the Safari as the default browser, then it works just as you described.

But given Edge doesn't do that on Windows for example, I don't think we can apply that to every platform. So maybe we can allow always opening a new tab in a spec level, per the platform convention?

annevk:
Okay, I've gone back to the initial version of giving implementations more freedom and added encouragement to match platform conventions.

notifications.bs
```diff
@@ -541,18 +557,53 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
+ <li>
+  <p>If <var>navigationURL</var> is non-null:
+
+  <ol>
+   <li>
+    <p>Select one of the following two options in an <a>implementation-defined</a> manner:
+
+    <ul>
+     <li><p><a for=/>Navigate</a> an existing <a for=/>top-level traversable</a> within the user
+     agent's <a for="user agent">top-level traversable set</a> to <var>navigationURL</var>.
```
saschanaz, line 585: But does Safari actually focus-and-navigate a tab? I think it just focuses a tab and does not navigate per what I see for the non-push external clicks. But I may be wrong.

Also does this need to say about focusing?

notifications.bs
```diff
@@ -541,18 +557,53 @@ interpreted as a language tag. Validity or well-formedness are not enforced. [[!
...
+     <!-- Should maybe set userInvolvement correctly here and below. -->
```
saschanaz, line 586: Probably set as "browser UI"?

notifications.bs
saschanaz: But it's for method for Navigation API case, which is different here? But navigateTo also sounds like a verb. Maybe it's fine anyway as most of the usage will be as a dictionary member than as an attribute.

(I still cannot unresolve the comment, but I'm not blocking on this anyway)

saschanaz (approved):
LGTM with some nonblocking comments.

notifications.bs
annevk: It definitely navigates for me when follow a link from a different app or run `https://github.com/whatwg/notifications/pull/213`. Just lost my initial comment trying that.

PR #227, opened 2024-12-05
Editorial: add "has showNotification() been successfully invoked"

annevk:
This is a concept needed for Declarative Web Push: see https://github.com/w3c/push-api/issues/360 for context.

***
<a href="https://whatpr.org/notifications/227.html" title="Last updated on Dec 5, 2024, 1:55 PM UTC (b1ce818)">Preview</a> | <a href="https://whatpr.org/notifications/227/d7f3e1a...b1ce818.html" title="Last updated on Dec 5, 2024, 1:55 PM UTC (b1ce818)">Diff</a>

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -1064,6 +1064,12 @@ partial interface ServiceWorkerGlobalScope {
 };
 </pre>
 
+<p>A <a for=/>service worker registration</a> has an associated
+<dfn export for="service worker registration">has <code>showNotification()</code> been successfully invoked</dfn>
+(a <a for=/>boolean</a>), which is initially false.
+
+<p class=note>This is infrastructure for the <cite>Push API</cite> standard. [[PUSH-API]]
+
 <div algorithm>
 <p>The
 <dfn method for=ServiceWorkerRegistration><code>showNotification(<var>title</var>, <var>options</var>)</code></dfn>
@@ -1097,6 +1103,10 @@ method steps are:
    <var>promise</var> with a {{TypeError}}, and abort these steps.
 
    <li><p>Run the <a for=/>notification show steps</a> for <var>notification</var>.
+
+   <li><p>Set <var>notification</var>'s <a for=notification>service worker registration</a>'s
+   <a for="service worker registration">has <code>showNotification()</code> been successfully invoked</a>
+   to true.
 
    <li><p><a>Queue a global task</a> on the <a>DOM manipulation task source</a> given
    <var>global</var> to <a for=/>resolve</a> <var>promise</var> with undefined.


notifications.bs
```diff
@@ -1098,6 +1104,10 @@ method steps are:
 
    <li><p>Run the <a for=/>notification show steps</a> for <var>notification</var>.
 
+   <li><p>Set <var>notification</var>'s <a for=notification>service worker registration</a>'s
+   <a for="service worker registration">has <code>showNotification()</code> been successfully invoked</a>
+   to true.
```
saschanaz, line 1109: Nothing sets it back to false, perhaps the push event handler should set it false in https://github.com/w3c/push-api/pull/385?

saschanaz: I also wonder what would happen if another push event is fired before the previous one finishes, because then this value would conflict. Perhaps we could have PushEvent.showNotification() (where the default arg is the DWP parse result) to make it per-event? 🤔

annevk: Right, I was waiting to see if this approach (which also means you have to process pushes sequentially) works for everyone, but then nobody gave feedback for what feels like an eternity…

annevk: @saschanaz it seems good to resolve this as well.

The idea here is indeed for the Push API PR to set this to false and for the Push API specification to enforce incoming messages are processed sequentially. (Perhaps scoped to a site, but I don't think we have to go into that much detail for now.)

saschanaz: I don't think Gecko waits for service worker until processing the next event, the [PushService notifies the service worker](https://searchfox.org/mozilla-central/rev/67c9592293e8d9f143e6053c8cfd6b7723207e3c/dom/push/PushService.sys.mjs#685-698) and moves on for the next push message. There should be something that ties a PushEvent and showNotification.

saschanaz: Does WebKit makes sure it's sequential btw? [queueTaskToFirePushEvent](https://searchfox.org/wubkat/rev/c46918d1cba365017d1df98e3d14f04db42320bc/Source/WebCore/workers/service/context/ServiceWorkerThread.cpp#290-317) looks like the queued task would end after adding a callback to the extended lifetime promise, is there an internal test for that?

annevk: I think the idea is to not queue another task until the callback is called. But I'll double check. (I don't think you can tie them together without API changes and additional complexity for web developers which doesn't really seem warranted and rather undesirable at this point. I investigated this pretty extensively a year ago or so.)

annevk: Brady confirmed they are indeed processed sequentially in that way.

saschanaz: I think I'm convinced, it's probably worth forcing the sequentialness here as we probably don't want to let thousands of push messages to run a huge number of parallel service worker events.

saschanaz: (Why can't I resolve my own comment? 🤦🏻‍♀️)

saschanaz (approved):

PR #232, opened 2025-02-28
Always limit the maximum number of actions to 2

saschanaz:
- [ ] At least two implementers are interested (and none opposed):
   * Gecko
   * …
- [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * … 
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: (N/A, actions are not supported yet)
   * WebKit: (N/A, actions are not supported yet)
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …
- [ ] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -144,12 +144,13 @@ support these features might ignore them.
  <dd>Null or a resource. It is initially null.
 </dl>
 
-<p>Users may activate actions, as alternatives to activating the notification itself. The
-<dfn>maximum number of actions</dfn> supported is an <a>implementation-defined</a> integer of zero
-or more, within the constraints of the notification platform.
+<p>Users may activate actions, as alternatives to activating the notification itself.
 
-<p class=note>Since display of actions is platform-dependent, developers are encouraged to make sure
-that any action an end user can invoke from a notification is also available within the web
+<p class=note>The maximum number of supported actions are limited to 2 on all platforms, to prevent
+fingerprinting based on the actual maximum number of the underlying system.
+
+<p class=note>Since the support of actions is platform-dependent, developers are encouraged to make
+sure that any action an end user can invoke from a notification is also available within the web
 application.
 
 <p class="note no-backref">Some platforms might modify an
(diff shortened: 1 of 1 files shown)


notifications.bs
```diff
@@ -258,8 +259,10 @@ important information through, e.g., loss of color or clipped corners.
  <li><p>Set <var>notification</var>'s <a for=notification>actions</a> to « ».
 
  <li>
-  <p>For each <var>entry</var> in <var>options</var>["{{NotificationOptions/actions}}"], up to the
-  <a>maximum number of actions</a> supported (skip any excess entries):
+  <p>Let |actions| be a [=list=] of initial maximum two items from
```
saschanaz, line 262: This is my best effort to make a sublist without actual corresponding infra term. Filed https://github.com/whatwg/infra/issues/666.

I wonder we should throw here, though.

annevk:
I think not having `maxActions` is an improvement, but I'm not sure we should make the call what the maximum is going to be forever. That seems too early days.

saschanaz:
Not sure we can remove maxActions without saying actual max value?

saschanaz:
So the options are:

1. have the spec say the specific number, or
2. have maxActions but define it as "implementation defined, but should not depend on system limitation" (so that authors would have a way to detect UA-specific differences instead of having to do UA sniffing), or
3. do not allow the callers to get any actual max value and still allow differences by system limitation (this would the most confusing option and would force authors to do heavy UA sniffing)

1 is my current preference for best simplicity.

saschanaz:
Mozilla is prototyping maxActions as constant in Nightly for now, still open to changes.

PR #244, opened 2026-03-28
Add top-level origin check

### LINKED ISSUE #177: Usage in cross-origin iframes
Both Chrome and Firefox have disallowed calling `Notification.requestPermission()` and (somewhat implicitly) `new Notification()` in cross-origin iframes. I'm not sure this is encoded explicitly anywhere in the spec.

In conversation with @annevk we said that for other permissions policy-controlled features this would be done through the Permissions API's request to use algorithm, which has a [step for checking "allowed to use" as per PP](https://w3c.github.io/permissions/#dfn-permission-state).

However, the Notifications spec does not define a Permissions Policy feature. This is because Notifications isn't really compatible with permission delegation, i.e. an origin that requests (push) notification permission would not expect the permission to be scoped to the embedded document.

For that reason, simply specifying a permissions policy integration with a permission name and a default allowlist, like many other specs do, seems like a bad choice, as it would allow developers to override the allowlist.

The simplest solution might be to just add checks in this spec that disallow usage in cross-origin iframes, though we could also consider supporting a "only deny, no override" option in PP.

annevk:
Fixes #177.

- [ ] At least two implementers are interested (and none opposed):
   * …
   * …
- [ ] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * … 
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: …
   * WebKit: …
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …
- [ ] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

### DIFF
--- a/notifications.bs
+++ b/notifications.bs
@@ -792,9 +792,12 @@ constructor steps are:
   <p>Run these steps <a>in parallel</a>:
 
   <ol>
-   <li><p>If the result of <a>getting the notifications permission state</a> is not
-   "<code>granted</code>", then <a>queue a task</a> to <a>fire an event</a> named
-   <code>error</code> on <a>this</a>, and abort these steps.
+   <li><p>If <a>this</a>'s <a>relevant settings object</a>'s
+   <a for="environment settings object">origin</a> is not <a>same origin</a> with <a>this</a>'s
+   <a>relevant settings object</a>'s <a for="environment">top-level origin</a> or the result of
+   <a>getting the notifications permission state</a> is not "<code>granted</code>", then
+   <a>queue a task</a> to <a>fire an event</a> named <code>error</code> on <a>this</a>, and abort
+   these steps.
 
    <li><p>Run the <a for=/>notification show steps</a> for <var>notification</var>.
   </ol>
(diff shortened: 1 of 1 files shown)


johannhof (approved):
This seems to match Chrome's behavior, thanks

annevk:
Do you happen to know if we have any test coverage for this?

saschanaz:
What is "top level" here? [HTML](https://html.spec.whatwg.org/#concept-environment-top-level-origin) says "implementation defined"... Is it supposed to check the top window's origin?

Gecko blocks any request if a third party iframe is involved at all - ABA is also blocked. Per our telemetry nobody is doing ABA permission request and so far nobody complained.

annevk:
It's implementation-defined for shared/service workers only so I guess your question is what should happen for shared workers since service workers are handled already? It's a good question and I'm not sure any of that is well-defined at the moment.

saschanaz:
There's a tentative test https://wpt.fyi/results/notifications/cross-origin-nested.tentative.https.sub.html?label=experimental&label=master&aligned

johannhof:
Interesting, Chrome should not fail these. I filed https://issues.chromium.org/498745938.

We can probably also make it non-tentative?

